Issue #1057, opened 2022-11-23
Crash in get_app_icon_html() with missing project UUID in plugin entry point URL


mikkonie:
Reported by @stolpeo and currently occurring in Kiosc.

If a site app creates a timeline event linked to a project (which is not a common case but entirely acceptable), we get a crash in `get_app_icon_html()`. This is because the template tag assumes the plugin entry point URL to have a `project` kwarg without checking the plugin type.

## How to Reproduce

1. Create a `ProjectEvent` with a `project` foreign key and a site app as `app` (see example below)
2. Navigate to the details view (=Project Overview) or Timeline view of the project you refer to in the event
3. Observe the crash

Example for creating the event in `shell_plus`:

```
from projectroles.plugins import get_backend_api
timeline = get_backend_api('timeline_backend')
User = get_user_model()
p = Project.objects.filter(type='PROJECT').first()
e = timeline.add_event(project=p, app_name='example_site_app', user=User.objects.first(), event_name='test_event', description='Test', status_type='OK')
```

## Spec

- If `app != 'projectroles'`, check for plugin type in addition to presence of `ProjectEvent.project` before assigning the project UUID URL kwarg
    * If not `ProjectAppPlugin`, do not except the project UUID URL kwarg
- Just to be safe, add try-except to `reverse()`
    * Set URL to `None` if we get an exception

## Tasks

- [ ] Add plugin type check
- [ ] Add try-except to `reverse()`
- [ ] Add test(s) for event linked to a project created by a site app plugin

gromdimon:
Seems to be done

PR #1070, opened 2022-12-08
get_app_icon_html and tests

gromdimon:
Task #1057 get_app_icon_html().
 - Changed a little bit timeline/templatetags/timeline_tags.py
 - Added 2 tests in timeline/tests/test_templatetsgs.py
 ---

Will be happy for any feedback

> Was: Screenshot from 2022-12-08 15-47-22
> Became: Screenshot from 2022-12-08 16-47-41

P.s.
I've also made changes on another tasks at my branch, so I hope, I do not push other files.

### DIFF
--- a/timeline/templatetags/timeline_tags.py
+++ b/timeline/templatetags/timeline_tags.py
@@ -6,6 +6,7 @@ from django.urls import reverse
 from django.utils.timezone import localtime
 
 from djangoplugins.models import Plugin
+from projectroles.plugins import ProjectAppPluginPoint
 
 from timeline.api import TimelineAPI
 from timeline.models import ProjectEvent
@@ -83,6 +84,10 @@ def get_app_icon_html(event, plugin_lookup):
         icon = ICON_PROJECTROLES
     else:
         plugin_name = event.plugin if event.plugin else event.app
+        if plugin_name in plugin_lookup.keys():
+            plugin = plugin_lookup[plugin_name]
+            if not isinstance(plugin, ProjectAppPluginPoint):
+                url_kwargs['project'] = None
         if plugin_name in plugin_lookup.keys():
             plugin = plugin_lookup[plugin_name]
             entry_point = getattr(plugin, 'entry_point_url_id', None)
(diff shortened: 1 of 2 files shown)


timeline/tests/test_templatetags.py
```diff
@@ -180,6 +180,39 @@ def test_get_app_icon_html_plugin(self):
...
+        url = reverse(
+            plugin.entry_point_url_id,
+            kwargs={'project': self.project.sodar_uuid},
+        )
+        ret = tags.get_app_icon_html(event, self.plugin_lookup)
+        self.assertIn(plugin.icon, ret)
+        self.assertIn(plugin.title, ret)
+        self.assertIn(url, ret)
+
+    def test_get_app_icon_html_timeline(self):
```
mikkonie: We can omit this testcase, as the timeline app itself is not expected to create any timeline events.

timeline/tests/test_templatetags.py
```diff
@@ -180,6 +180,39 @@ def test_get_app_icon_html_plugin(self):
...
+        )
+        plugin = get_app_plugin('example_site_app')
+        url = reverse(
+            plugin.entry_point_url_id,
+            kwargs={'project': self.project.sodar_uuid},
+        )
+        ret = tags.get_app_icon_html(event, self.plugin_lookup)
+        self.assertIn(plugin.icon, ret)
+        self.assertIn(plugin.title, ret)
+        self.assertIn(url, ret)
```
mikkonie, line 200: We should also assert that `url` corresponds to the url reversed from the `entry_point_url_id` of the plugin. Other than that the test case looks very good.

timeline/templatetags/timeline_tags.py
```diff
@@ -86,8 +86,15 @@ def get_app_icon_html(event, plugin_lookup):
         if plugin_name in plugin_lookup.keys():
             plugin = plugin_lookup[plugin_name]
             entry_point = getattr(plugin, 'entry_point_url_id', None)
-            if entry_point:
-                url = reverse(entry_point, kwargs=url_kwargs)
+            if entry_point in plugin_lookup.items:
```
mikkonie, line 94: This try-except is good as it fixes the immediate problem of the crash.

We still need to have the check for the plugin type, as specified in the issue. ONLY provide `project` in `url_kwargs`, if the if the plugin is based on `ProjectAppPlugin`. Site and backend apps will not have the project kwarg in their entry point URL.

gromdimon:
I've just changed two files and wanted to update this pull request, but I closed it somehow )(
I hope, it's allright

Issue #1055, opened 2022-11-21
Timeline modal fails for extra data after viewing event details


mikkonie:
### Describe the Bug

In the `timeline` app, after opening the modal for event details, the same content remains visible when closing and reopening the modal for extra data.

### How to Reproduce

1. Open the modal for event details (via the link in event timestamp)
2. Close the modal
3. Open the modal for event extra data (via the extra data link on the right hand side of the event list)

### Expected Behavior

Extra data should be visible in the modal content instead of the event details.

### Screenshots

![timeline](https://user-images.githubusercontent.com/11211145/204860917-2ce48abc-83b2-4a6f-8dcc-4a23c1ef0b04.png)

gromdimon:
Should be closed

PR #1072, opened 2022-12-20
Fix timeline issues (#1057, #1076, #1081)

gromdimon:
I decided not to delete previous versions of realisation of extra_data modal, but just to comment them.


.gitignore
```diff
@@ -100,3 +100,6 @@ flamegraph.pl
 
 # Celery
 celerybeat-schedule
+
+# Questions
+Q.md
```
mikkonie: This doesn't seem to be any standard file we use. Consider removing from commit.

timeline/models.py
```diff
@@ -154,6 +156,19 @@ def get_status_changes(self, reverse=False):
             '{}pk'.format('-' if reverse else '')
         )
 
+    def get_status_changes_json(self):
```
mikkonie: I have some questions/issues with this method:

1. It's called `get_status_changes_json()`, but we already handle the status changes in the event details modal.
2. It seems this method somehow tries to dump the entire model as JSON?
    - What we need to return for the extra data modal is the JSON data from the `extra_data` fields in the `ProjectEvent` object and its related `ProjectEventStatus` objects, if present.
    - Hint: see `collect_extra_data()` for an example how this has been done on the server side.

timeline/static/timeline/js/extra_data.js
```diff
@@ -0,0 +1,32 @@
+$(document).ready(function() {
```
mikkonie: This code can be moved within timeline.js, I believe there is no practical reason for loading multiple JQuery files for the timeline templates.

timeline/templates/timeline/timeline.html
```diff
@@ -72,6 +72,9 @@ <h3><i class="iconify" data-icon="mdi:clock-time-eight"></i> {{ timeline_title }
   <!-- Event list -->
   <script type="text/javascript" src="{% static 'timeline/js/timeline.js' %}"></script>
 
+  <!-- Extra data -->
+  <script type="text/javascript" src="{% static 'timeline/js/extra_data.js' %}"></script>
```
mikkonie: See earlier comment, this JQuery can be moved into timeline.js.

timeline/templatetags/timeline_tags.py
```diff
@@ -85,9 +86,19 @@ def get_app_icon_html(event, plugin_lookup):
         plugin_name = event.plugin if event.plugin else event.app
         if plugin_name in plugin_lookup.keys():
             plugin = plugin_lookup[plugin_name]
+            if not isinstance(plugin, ProjectAppPluginPoint):
+                url_kwargs['project'] = None
```
mikkonie: This looks like it should work fine :) However, there's still one small thing to refactor. Right now, you may first set the project UUID and then reset it to None (see line 76), which is a bit redundant.

If the app is projectroles, you can simply add the project as a kwarg directly to the `reverse()` call on line 82 and set up the `url_kwargs` variable later, if we find a plugin on line 88.

timeline/tests/test_templatetags.py
```diff
@@ -180,6 +180,21 @@ def test_get_app_icon_html_plugin(self):
...
+            project=self.project,
+            app='example_site_app',
+            user=self.user_owner,
+            event_name=TEST_EVENT_NAME,
+            description=TEST_EVENT_DESC,
+        )
+        ret = tags.get_app_icon_html(event, self.plugin_lookup)
+        self.assertIn(plugin.icon, ret)
+        self.assertIn(plugin.title, ret)
+        self.assertIn('', ret)
```
mikkonie: This just asserts that an empty string is included in `ret`, probably returns true every time :)
I also recommend adding an assertion for the correct entry point URL being returned in the result, similar to what is done in ` test_get_app_icon_html_plugin()`.

mikkonie (commented):
A few general comments:

- Please refer to the issue(s) involved, at least by ID, in your pull request title. That makes it easier for tracking.
- If at all possible, I'd prefer you to have your work branches in the official repo in the future. That makes it easier for e.g. me to quickly pull a branch to test it locally, without having to track more remotes. As a contributor you should have a right to push your own work branches.

.gitignore
gromdimon: That was a file, where I took notes and questions. Now I moved it into another directory and deleted from .gitignore.

timeline/static/timeline/js/extra_data.js
gromdimon: Understood. Since I am not familiar with jquery, I can make inefficcient solutions( Sorry about that...

timeline/templatetags/timeline_tags.py
gromdimon: In that way it is also possible to set up `url_kwargs` variable not on line 88, but later (in new version of file it is line 89) in another `reverse()` for `entry_point`.

timeline/tests/test_templatetags.py
gromdimon: I have to say, that I am confused here( 
The thing is, that `reverse(entry_point, kwargs={'project':None})` raises exception => the url is being setted up to `''`. So in ret there will be an html without any `href=`. 
Maybe I misunderstood something((

gromdimon:
Sorry for inconveniences. I will make a branch in the official repo and make changes there. Secondly, this two tasks reference to #1055 and #1057 . I will make better next time!

timeline/tests/test_templatetags.py
mikkonie: Note that `get_app_icon_html()` returns the entire HTML as string and you store that in `ret`. Hence, you're just asserting that an empty string is included as a substring of the full string, which returns true every time. You can test this out by replacing `ret` with any random string :)

Instead, you should e.g. assert that `href=` does not appear in the returned HTML.

timeline/templatetags/timeline_tags.py
```diff
@@ -87,7 +85,15 @@ def get_app_icon_html(event, plugin_lookup):
             plugin = plugin_lookup[plugin_name]
             entry_point = getattr(plugin, 'entry_point_url_id', None)
             if entry_point:
-                url = reverse(entry_point, kwargs=url_kwargs)
+                try:
+                    url = reverse(entry_point, kwargs={'project':None})
```
mikkonie: There is a problem with this implementation. In this code we **only** use the project UUID for the URL if the event app is `projectroles`. The UUID should also be given if the app plugin is of type `ProjectAppPlugin`, as I previously explained.

You should also ensure the unit tests catch this, so make sure you have tests for events where the app is 1) projectroles, 2) some project app and 3) some non-project app. Explicitly assert the reversed URL based on the expected `entry_point_url_id` of the plugin is returned. For projectroles, assert the URL is `projectroles:detail` instead of course.

timeline/views_ajax.py
```diff
@@ -48,6 +50,116 @@ def get_event_details(self, event):
         return ret
 
 
+class ExtraDataHTMLMixin:
```
mikkonie: This looks good. I would suggest to place the content of this mixin as protected methods under `EventExtraMixin`. If we find more use cases for JSON formatting in the future, we might refactor these into a general helper in projectroles, but for now let's keep it simple :)

timeline/views_ajax.py
```diff
@@ -12,6 +13,7 @@
 
 from timeline.models import ProjectEvent
 from timeline.templatetags.timeline_tags import get_status_style
+from timeline.templatetags.timeline_tags import get_event_extra_data, collect_extra_data
```
mikkonie: If there are methods we now use only in client side rendering instead of server side template tags, make sure to move them away from the template tag module under an appropriate view or mixin.

timeline/views_ajax.py
```diff
@@ -48,6 +50,116 @@ def get_event_details(self, event):
...
+            str_list.append('\n')
+
+        str_list.append('</span>')
+        str_list.append('  ' * (indent - 1))
+        str_list.append('<span class="json-close-bracket">]</span>')
+
+class EventExtraMixin(ExtraDataHTMLMixin):
+    """Mixin for event extra data retrieval helpers"""
+
+    def form_html(self, event):
```
mikkonie: One issue I noticed: it looks like indentation isn't working in the displayed JSON.

Also, since this is structured data I'd prefer to see it displayed using a monospace font (`font-family: monospace` would conform to SODAR Core style).

2023-01_extra

gromdimon: I implemented all this changes except the last obe with font and idents. Have no idea, why jquery just ignores them , but I will try to find something out!

timeline/templatetags/timeline_tags.py
```diff
@@ -70,24 +69,35 @@ def get_plugin_lookup():
...
+                'projectroles:detail',
+                kwargs={'project': event.project.sodar_uuid},
+            )
         title = 'Projectroles'
         icon = ICON_PROJECTROLES
     else:
         plugin_name = event.plugin if event.plugin else event.app
         if plugin_name in plugin_lookup.keys():
+            url_kwargs = {}
+            if event.project:
```
mikkonie: Here, instead of `event.project` we just have to check for the app type. As I wrote in teams, you can do this as follows: `isinstance(your_plugin_here, ProjectAppPluginPoint)`

Aslo, make sure your tests catch this.

timeline/templatetags/timeline_tags.py
```diff
@@ -70,24 +69,35 @@ def get_plugin_lookup():
...
     """Return icon link HTML for app by plugin lookup"""
     url = None
-    url_kwargs = {}
-    if event.project:
-        url_kwargs['project'] = event.project.sodar_uuid
     title = event.app
     icon = ICON_UNKNOWN_APP  # Default in case the plugin is not found
 
     if event.app == 'projectroles':
         if event.project:
```
mikkonie, line 77: This condition can be omitted, as the projectroles entry point URL is fixed and always has the project UUID as kwarg. Also, projectroles will not create events without an associated project.

timeline/tests/test_templatetags.py
```diff
@@ -180,6 +180,21 @@ def test_get_app_icon_html_plugin(self):
         self.assertIn(plugin.title, ret)
         self.assertIn(url, ret)
```
mikkonie, line 181: As said before, check the URLs in these tests explicitly to ensure we get the appropriate URL. Do this with:
1. Event created with `app=projectroles`
2. Event created with some other project app (`example_project_app` is a good choice)
3. Event created with a site app (`example_site_app` is recommended :))

Then assert the URL returned corresponds to whatever `reverse()` returns with the related `entry_point_url_id` and valid kwargs. In case of projectroles the entry point is hardcoded to `projectroles:detail`.

timeline/views_ajax.py
```diff
@@ -65,6 +169,27 @@ def get(self, request, *args, **kwargs):
         return Response(self.get_event_details(event), status=200)
 
 
+class ProjectEventExtraAjaxView(EventExtraMixin, SODARBaseProjectAjaxView):
+    """Ajax view for retrieving event extra data for projects"""
+
+    permission_required = 'timeline.view_timeline'
+    allow_anonymous = True
```
mikkonie: `allow_anonymous` should not be set true in these views. This is not your fault as it seems to be incorrect in the existing event detail ajax views as well. Feel free to remove.

Also, please add permission tests for these new ajax views under `timeline.tests.test_permissions_ajax`.

gromdimon:
Made some changes

timeline/templatetags/timeline_tags.py
gromdimon: Implemented this

timeline/templatetags/timeline_tags.py
gromdimon: Implemented this as well

timeline/views_ajax.py
gromdimon: Deleteted this line

Issue #1059, opened 2022-11-30
User name for initial owner in project create form not following regular convention


mikkonie:
### Describe the Bug

When creating a project, the initial owner's name is not presented with the standard syntax of the user dropdown. Minor issue, but looks confusing.

### How to Reproduce

1. Navigate to a category under which you are able to create subcategories or projects
2. Click on "create project or category"
3. Observe the initial name in  the "owner" field

### Expected Behavior

This should be given in the same format provided by the dropdown when querying for choices: `Full Name (username) <email@example.com>`.

### Tasks

- [ ] Update initial value
- [ ] Assert the initial value in `projectroles.tests.test_views.TestProjectCreateView`

mikkonie:
Starting points:

- The initial values for the widget are set in the following 4 lines: https://github.com/bihealth/sodar-core/blob/dev-0.11.1/projectroles/forms.py#L508
- The formatting of the user name is done here: https://github.com/bihealth/sodar-core/blob/dev-0.11.1/projectroles/views_ajax.py#L459

How to do this:

- Move the `get_result_label()` implementation under the `SODARUser` model as `get_form_label()`
    * Remember to add test(s) for this in `example_site.users.tests.test_models`
- Return `user.get_form_label()` in the existing `get_result_label()` method
    * This method must still be implemented for django-autocomplete-light, so don't remove it
- In `ProjectForm`, after setting `self.initial['owner']`, set the following:
    * `self.fields['owner'].label_from_instance = lambda x: x.get_form_label()`
    * You only need to call this once on line 512

gromdimon:
Thank you very much for help!
I did:
- [x] `get_form_label()` method in `SODARUser` model
- [x] Added test
- [x] Changed existing implementation to  `user.get_form_label()`
- [ ] But I did not change asssert in the initial value in `projectroles.tests.test_views.TestProjectCreateView` as there is another logic without this function

gromdimon:
Will be happy for any comments

mikkonie:
Done by @gromdimon and merged.

PR #1092, opened 2023-01-12
Fix/owner name

gromdimon:
Fix of #1059 

- [x] Updated initial value (in 2 places)
- [x] Wrote a test for function


projectroles/forms.py
```diff
@@ -507,8 +507,14 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
             # Set owner
             if self.current_user.is_superuser and parent_project:
                 self.initial['owner'] = parent_project.get_owner().user
+                self.fields[
```
mikkonie: We only need to set the label once after the if/else clause, as it's the same code regardless of which initial value we use.

projectroles/models.py
```diff
@@ -1129,6 +1129,15 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
         self.set_group()
 
+    def get_form_label(self):
+        """Return options with name, username and email"""
+        display = '{}{}{}'.format(
```
mikkonie: This can be simplified into simply returning the string instead of first setting it into the `display` variable. No idea why we did that in the original implementation :)

mikkonie (changes requested):

mikkonie (approved):

PR #1093, opened 2023-01-12
Create/admin site

### LINKED ISSUE #1081: Add UI tests for timeline event details and extra data
When re-implementing timeline extra data retrieval, old UI tests were removed as they were no longer applicable. Meanwhile, event detail UI tests never existed in the first place.

We should add some sanity checks in UI tests to ensure the data for both event detail and extra data modals gets correctly displayed.

I recommend doing this at the same time or after #873.

## Tasks

- [x] Add tests to assert event details are correctly displayed
- [x] Add tests to assert event extra data is correctly displayed
- [x] Add tests to assert extra data clipboard copying works

gromdimon:
Init PR


timeline/plugins.py
```diff
@@ -114,3 +114,28 @@ class SiteAppPlugin(SiteAppPluginPoint):
 
     #: Required permission for displaying the app
     app_permission = 'timeline.view_site_timeline'
+
+
+class AdminSiteAppPlugin(SiteAppPluginPoint):
+    """Projectroles plugin for registering the app"""
+
+    #: Name (slug-safe, used in URLs)
+    name = 'admin_site'
```
mikkonie: Plugin names are global, so we should be careful with these and follow a certain hierarchy and conventions. The other plugins within `timeline.plugins` are named as follows:

- `timeline`
- `timeline_backend`
- `timeline_site`

Hence the logical name would be `timeline_site_admin`.

timeline/plugins.py
```diff
@@ -114,3 +114,28 @@ class SiteAppPlugin(SiteAppPluginPoint):
...
+
+    #: Title (used in templates)
+    title = 'Admin Site Events'
+
+    #: App URLs (will be included in settings by djangoplugins)
+    urls = urlpatterns
+
+    #: Iconify icon
+    icon = 'mdi:clock-time-eight'
```
mikkonie: I suggest choosing a different icon for this to differentiate it from the site-wide events view. `mdi:clock-time-eight` and `mdi:clock-time-eight-outline` are already in use elsewhere by the timeline app, so how about going with some other clock icon in Material Design Icons? You can browse them here: https://materialdesignicons.com/

timeline/plugins.py
```diff
@@ -114,3 +114,28 @@ class SiteAppPlugin(SiteAppPluginPoint):
...
+
+    #: Description string
+    description = 'Timeline of Admin Site Events'
```
mikkonie: Maybe "Admin view for all site events" would be more descriptive here.

timeline/plugins.py
```diff
@@ -114,3 +114,28 @@ class SiteAppPlugin(SiteAppPluginPoint):
...
+
+
+class AdminSiteAppPlugin(SiteAppPluginPoint):
+    """Projectroles plugin for registering the app"""
+
+    #: Name (slug-safe, used in URLs)
+    name = 'admin_site'
+
+    #: Title (used in templates)
+    title = 'Admin Site Events'
```
mikkonie: "All Timeline Events" would probably be more descriptive here.

timeline/templates/timeline/admin_site.html
```diff
@@ -0,0 +1,106 @@
...
+  {{ block.super }}
+
+  <!-- Event list -->
+  <script type="text/javascript" src="{% static 'timeline/js/timeline.js' %}"></script>
+
+  <!-- Tour content -->
+  <script type="text/javascript">
+      tourEnabled = true;
+
+      tour.addStep('event_list', {
```
mikkonie: This tour help message is, of course, not relevant for the admin view, but we can unify these once we go ahead with merging the templates (if feasible).

timeline/urls.py
```diff
@@ -19,6 +19,11 @@
         view=views.SiteTimelineView.as_view(),
         name='list_site',
     ),
+    url(
+        regex=r'^admin_site$',
```
mikkonie: In URLs we should also follow conventions closely, but I haven't really documented these properly so my bad :)

I would recommend changing this into `^site/all$`. We generally want to avoid underscores in URL slugs and the string "admin" can have some funny consequences (I've learned this the hard way..)

timeline/views.py
```diff
@@ -53,6 +53,19 @@ def get_queryset(self):
         return ProjectEvent.objects.filter(**set_kwargs).order_by('-pk')
 
 
+class EventAdminTimelineMixin:
```
mikkonie: If this is only used in `AdminTimelineView`, the methods can go directly under that view class.

timeline/plugins.py
gromdimon: - [x] Changed

timeline/plugins.py
gromdimon: I chose an cog-counterclockwise icon. Thoughts were "rollback for debug"

timeline/plugins.py
gromdimon: - [x] Changed

timeline/plugins.py
gromdimon: - [x] Changed

timeline/templates/timeline/admin_site.html
gromdimon: - [x] Deleted it

gromdimon: However, will reform the whole template later as it was discussed

timeline/urls.py
gromdimon: - [x] Changed

timeline/views.py
gromdimon: - [x] Moved

gromdimon:
Added badge into description

projectroles/forms.py
```diff
@@ -507,8 +507,14 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
             # Set owner
             if self.current_user.is_superuser and parent_project:
                 self.initial['owner'] = parent_project.get_owner().user
+                self.fields[
+                    'owner'
+                ].label_from_instance = lambda x: x.get_form_label()
             else:
                 self.initial['owner'] = self.current_user
+                self.fields[
```
mikkonie: Note that this branch still contains this older version of the user dropdown fix implementation. We don't want to override the existing already merged solution with this one.

timeline/templatetags/timeline_tags.py
```diff
@@ -37,6 +37,13 @@ def get_timestamp(event):
     return localtime(event.get_timestamp()).strftime('%Y-%m-%d %H:%M:%S')
 
 
+@register.simple_tag
```
mikkonie: A template tag is not needed, you can simply call `user.is_superuser` in a template.

timeline/templates/timeline/_list_item.html
```diff
@@ -31,6 +31,19 @@
...
   <td>
+    {% check_superuser user as perm %}
+    {% if event.project and perm and site_admin %}
+        <span class="badge"
+              title="{{ event.project }}"
+              data-toggle="tooltip" data-placement="top">
+        <i class="iconify"
+             data-icon="{% if event.project.type == 'PROJECT' %}mdi:cube{% else %}mdi:rhombus-split{% endif %}">
+        </i>
+        <a href="{% url 'projectroles:detail' project=event.project.sodar_uuid %}">
```
mikkonie: Please format templates with a 2-space indent. As an exception to this, when splitting a long tag on multiple lines, it's ok to place the attributes starting in the same position for clarity. E.g. above here in the icon, "class" and "data-icon" would be aligned on the same position.

projectroles/forms.py
gromdimon: I deleted all previous changes concerning owner name bug. However I do not understand, why do the github showed changes between branches. Dev is now merged with fix/owner-name )(

timeline/templatetags/timeline_tags.py
gromdimon: I reinvented a bicycle :) Fixed

timeline/templates/timeline/_list_item.html
gromdimon: I definetely should learn more about html formating! Thanks, fixed

gromdimon:
Made another changes. May I start working on unifying templates?

timeline/tests/test_permissions.py
```diff
@@ -59,6 +59,21 @@ def test_stite_list_anon(self):
...
+        url = reverse('timeline:timeline_site_admin')
+        good_users = [
+            self.superuser,
+        ]
+        self.assert_response(url, good_users, 200)
+        self.assert_response(url, self.anonymous, 302)
+
+    @override_settings(PROJECTROLES_ALLOW_ANONYMOUS=True)
+    def test_admin_list_anon(self):
+        """Test admin list with anonymous assess"""
```
mikkonie: Spotted a typo in "access" ;)

timeline/tests/test_permissions.py
```diff
@@ -59,6 +59,21 @@ def test_stite_list_anon(self):
         url = reverse('timeline:list_site')
         self.assert_response(url, self.anonymous, 302)
 
+    def test_admin_list(self):
+        """Test admin event list"""
+        url = reverse('timeline:timeline_site_admin')
+        good_users = [
```
mikkonie, line 65: Bad users are missing. It iks very important to test for all user roles.

timeline/views.py
```diff
@@ -35,9 +35,11 @@ def get_context_data(self, *args, **kwargs):
                 get_display_name(project.type, title=True)
             )
             context['timeline_mode'] = 'project'
+            context['event_type'] = 'project'
         else:
```
mikkonie: `timeline_mode` already handles this, no?

timeline/views.py
```diff
@@ -78,7 +80,30 @@ class SiteTimelineView(
...
+class AdminTimelineView(
+    LoginRequiredMixin,
+    LoggedInPermissionMixin,
+    ListView,
+):
+    """View for displaying timeline events for admin site view"""
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context['timeline_title'] = 'Admin site Events'
```
mikkonie: This should also be renamed to "All Timeline Events"

timeline/templates/timeline/timeline.html
```diff
@@ -1,8 +1,9 @@
-{% extends 'projectroles/project_base.html' %}
```
mikkonie: These edits to `timeline.html` are incorrect. As we agreed yesterday in Teams, this file should remain as is, as it is included in a project view and uses a different base template to site views.

The templates for the site-wide event view and admin view should be unified, as they use the same base template and should be 99% identical.

We should end up with two templates:

- `timeline.html` used for project view
    * Based on `project_base.html`, content under the `projectroles_extend` block
    * In other words, no changes to prior implementation
- `timeline_site.html`
    * Used for both site-wide event view and the admin view
    * Should only require slight edits to the current `timeline_site.html`
    * Most likely won't require much changes apart from than title/heading?
    * I noticed the existing template has erroneously some project access references included, please do remove these as we, by definition, never have a project context in a site app view

timeline/tests/test_views.py
```diff
@@ -193,3 +193,33 @@ def test_render(self):
...
+                reverse(
+                    'timeline:timeline_site_admin',
+                )
+            )
+            self.assertEqual(response.status_code, 200)
+            self.assertEqual(len(response.context['object_list']), 2)
+            self.assertEqual(
+                response.context['object_list'].first(), self.event_site
+            )
+            self.assertEqual(response.context['object_list'][1], self.event)
```
mikkonie: These are looking good. Please also add UI test(s) for the admin view to assert we are seeing the project title badges in events. This is a good sanity check to assure we are rendering things correctly.

timeline/templates/timeline/timeline_site.html
```diff
@@ -1,106 +0,0 @@
-{% extends 'projectroles/base.html' %}
```
mikkonie: This template should not be deleted, instead it should form the basis for both the site-wide events and the admin view.

timeline/templates/timeline/_list_item.html
```diff
@@ -31,6 +31,16 @@
     {% endif %}
   </td>
   <td>
+    {% if event.project and user.is_superuser and site_admin %}
+      <span class="badge" title="{{ event.project }}" data-toggle="tooltip" data-placement="top">
```
mikkonie: The `badge-info` class I requested is still missing, please add.

timeline/templates/timeline/timeline.html
mikkonie: For differences in project app and site app templates, see here:

- [Project app templates](https://sodar-core.readthedocs.io/en/dev/dev_project_app.html#template-structure)
- [Site app templates](https://sodar-core.readthedocs.io/en/dev/dev_site_app.html#templates)

timeline/templates/timeline/timeline.html
```diff
@@ -1,8 +1,9 @@
...
+{% extends 'projectroles/base.html' %}
```
mikkonie: I'm still seeing changes to `timeline.html`. This template should not be modified in this pull request.

timeline/templates/timeline/side_timeline.html
```diff
@@ -6,8 +6,7 @@
 {% load projectroles_common_tags %}
```
mikkonie: The file name of this template is not consistent with SODAR Core conventions. It should remain `timeline_site.html`.

timeline/templates/timeline/side_timeline.html
```diff
@@ -17,19 +16,33 @@
...
 <div class="row sodar-subtitle-container bg-white sticky-top">
   <h2><i class="iconify" data-icon="mdi:clock-time-eight"></i> {{ timeline_title }}</h2>
   {% if timeline_mode == 'object' %}
     <div class="ml-auto">
-      <a href="{% url 'timeline:list_site' %}"
-         class="btn btn-secondary ml-auto" role="button">
-        <i class="iconify" data-icon="mdi:arrow-left-circle"></i>
-        Site-Wide Events
-      </a>
+      {% if timeline_modef == 'site' %}
```
mikkonie: Looks like there's a typo in checking the `timeline_mode` variable.

timeline/templates/timeline/side_timeline.html
```diff
@@ -45,7 +58,11 @@ <h2><i class="iconify" data-icon="mdi:clock-time-eight"></i> {{ timeline_title }
           </thead>
           <tbody>
             {% for event in object_list %}
-              {% include 'timeline/_list_item.html' %}
+              {% if timeline_mode == 'admin' %}
```
mikkonie: We should simply check `timeline_mode` in the included template instead. Since the variable comes from the view context, `_list_item.html` should be able to access it directly without any extra steps needed.

timeline/templates/timeline/side_timeline.html
```diff
@@ -55,9 +72,16 @@ <h2><i class="iconify" data-icon="mdi:clock-time-eight"></i> {{ timeline_title }
       {% include 'projectroles/_pagination.html' with pg_small=False %}
     {% endif %}
    {% else %}
-    <div class="alert alert-info" role="alert">
-      No site-wide timeline events found{% if timeline_mode == 'object' %} for this object{% endif %}.
-    </div>
+    {% if timeline_mode == 'site' %}
+      <div class="alert alert-info" role="alert">
+        No site-wide timeline events found{% if timeline_mode == 'object' %} for this object{% endif %}.
```
mikkonie: It seems this could be shortened into simply having the condition block around the string "site-wide".

timeline/tests/test_ui.py
```diff
@@ -188,3 +190,137 @@ def test_render_object(self):
...
+        self.login_and_redirect(
+            self.superuser, url, wait_elem=None, wait_loc='ID'
+        )
+        button = self.selenium.find_element(
+            By.CLASS_NAME, 'sodar-tl-link-detail'
+        )
+        button.click()
+        self.assertIsNotNone(
+            self.selenium.find_element(By.CLASS_NAME, 'sodar-card-table')
+        )
```
mikkonie, line 316: Generally, these UI tests look very good. Here, you should still check for the existence of the table rows that match the detais of the event.

timeline/tests/test_ui.py
```diff
@@ -188,3 +190,137 @@ def test_render_object(self):
...
+        self.login_and_redirect(
+            self.superuser, url, wait_elem=None, wait_loc='ID'
+        )
+        button = self.selenium.find_element(
+            By.CLASS_NAME, 'sodar-tl-link-extra-data'
+        )
+        button.click()
+        self.assertIsNotNone(
+            self.selenium.find_element(By.CLASS_NAME, 'modal-title')
+        )
```
mikkonie, line 358: Same as above, you should still assert the visibility of the extra data itself. Just asserting the visibility of the element containing the JSON is enough, we can trust that the content is rendered in that case as well.

timeline/tests/test_views.py
```diff
@@ -193,3 +193,33 @@ def test_render(self):
...
+
+    def test_render(self):
+        """Test rendering the admin list view"""
+        with self.login(self.user):
+            response = self.client.get(
+                reverse(
+                    'timeline:timeline_site_admin',
+                )
+            )
+            self.assertEqual(response.status_code, 200)
```
mikkonie: One small hint: when we do a request with `with self.login()`, the only rows later which must be contained within the `with` block are ones for asserting response redirecting (as it does some stuff which requires the user to be logged in). If you want to avoid unnecessary indentation, you can just de-indent the rest of these rows :)

timeline/templates/timeline/_list_item.html
```diff
@@ -31,6 +31,16 @@
     {% endif %}
   </td>
   <td>
+    {% if event.project and user.is_superuser and site_admin %}
+      <span class="badge badge-info mr-1" title="{{ event.project }}" data-toggle="tooltip" data-placement="top">
+        <i class="iconify"
+           data-icon="{% if event.project.type == 'PROJECT' %}mdi:cube{% else %}mdi:rhombus-split{% endif %}">
+        </i>
+        <a href="{% url 'projectroles:detail' project=event.project.sodar_uuid %}">
```
mikkonie: The `a` tag requires the `text-white` class here, otherwise we get a blue link on blue background.

timeline/templates/timeline/side_timeline.html
```diff
@@ -6,8 +6,7 @@
 {% load projectroles_common_tags %}
 
 {% block title %}
-  {{ timeline_title }} {% if timeline_mode == 'project' %}for{% else %}in{% endif %}
-  {{ project.title }}
+  {{ timeline_title }} in {{ project.title }}
```
mikkonie: The project title should be omitted here, as we never have a project context in a site view.

mikkonie (approved):

Issue #1096, opened 2023-01-13
Add event status extra data to event details modal


mikkonie:
Timeline `ProjectEventStatus` objects also have their own `extra_data` field, which can be used to store extra data for a specific status update in the event lifecycle.

In the previous implementation these were loaded together into the server-side modal we got rid of. In the new implementation, we are not yet displaying these. IMHO the most logical place for these to go UI-wise is the event detail modal, on the right hand side of the "description" field similar to the event list.

**NOTE:** This should be done as a separate pull request from the timeline tasks under work at the time of writing.

## Spec

- If status object has extra data, add extra data link to the status in the event details modal
- Once clicked, the modal should:
    * Display `sodar-modal-wait`
    * Query for extra data from an Ajax view
        + Probably it's easier to re-use the existing extra data ajax view for this, just allow it to also return extra data for a `ProjectEventStatus`
    * Once data is returned, render extra data modal content with the status extra data

## Tasks

- [ ] Add `ProjectEventStatus` extra data retrieval in view
- [ ] Implement logic in modal
- [ ] Add view, permission and UI tests

mikkonie:
Done by @gromdimon, closing.

PR #1106, opened 2023-01-20
Create/status extra data

gromdimon:
Issue #1096 
TODO:
 - [x] change jquery `.sodar-tl-link-extra-data-status` to something workable
 - [x] Add tests

### DIFF
--- a/timeline/static/timeline/js/timeline.js
+++ b/timeline/static/timeline/js/timeline.js
@@ -1,3 +1,25 @@
+var extra_data = function () {
+        var dataUrl = $(this).attr('data-url');
+        $('#sodar-modal').modal('hide');
+        $('#sodar-modal-wait').modal('show');
+        $.ajax({
+            url: dataUrl,
+            method: 'GET',
+        }).done(function (data) {
+            $('.modal-title').text(
+                'Event Extra Data: ' + data['app'] + '.' + data['name'] + ' (' +
+                data['user'] + ' @ ' + data['timestamp'] + ')');
+            $('.modal-body').html('').append($('<p>').html(
+                '<pre class="sodar-tl-json" id=data-to-clipboard>' + data['extra'] + '</pre>'));
+            $('#sodar-modal-wait').modal('hide');
+            $('#sodar-modal').modal('show');
+        }).fail(function (data) {
+            console.dir(data);
+            $('.modal-body').html('Error: ' + data.statusText);
+            $('#sodar-modal-wait').modal('hide');
+            $('#sodar-modal').modal('show');
+        });
+    };
 $(document).ready(function() {
     $('.sodar-tl-link-detail').click(function () {
         $('#sodar-modal-wait').modal('show');
@@ -22,14 +44,21 @@ $(document).ready(function() {
             );
             var tableBody = $('.modal-body').find('tbody');
             for (var i = 0; i < data['status'].length; i++) {
+                let editionalDescription = '<a class="sodar-tl-link-status-extra-data text-primary pull-right" ' +
+                    'data-url="' + data['status'][i]['extra_status_link'] + '"><i class="iconify" ' +
+                    'data-icon="mdi:text-box" title="Status Extra Data" data-toggle="tooltip" data-placement="right">' +
+                    '</i></a>';
+                if (data['status'][i]['no_extra_status'] === true) {
+                    editionalDescription = '';
+                }
                 tableBody.append($('<tr>')
                     .append($('<td>').html(data['status'][i]['timestamp']))
-                    .append($('<td>').html(data['status'][i]['description']))
+                    .append($('<td>').html(data['status'][i]['description'] + editionalDescription)
                     .append($('<td>')
                         .attr('class', data['status'][i]['class'])
                         .html(data['status'][i]['type'])
                     )
-                );
+                ));
             }
             $('#sodar-modal-wait').modal('hide');
             $('#sodar-modal').modal('show');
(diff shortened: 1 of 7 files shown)


timeline/static/timeline/js/timeline.js
```diff
@@ -61,3 +61,25 @@ $(document).ready(function() {
         });
     });
 });
+$('body').on('click', '.sodar-tl-link-status-extra-data', function () {
```
mikkonie: There is now quite a bit of repetition between clicking the extra data link in two different places. Please unify the repeating bits under e.g. a separate helper function and call that instead.

timeline/views_ajax.py
```diff
@@ -36,13 +36,30 @@ def get_event_details(self, event):
             'status': [],
         }
         status_changes = event.get_status_changes(reverse=True)
-        for s in status_changes:
+        for idx, s in enumerate(status_changes):
+            desc = ''
+            if s.extra_data:
+                desc = (
+                    '<a class="sodar-tl-link-status-extra-data text-primary pull-right" data-url="{}">'
```
mikkonie: We should form this HTML in the JQuery instead. We generally don't want to push preformatted HTML data to the client side in Ajax view.

The JSON formatting was an exception since it was really the best approach, but in the case of this simple link we should be able to easily create it in JQuery. Just returning the URL should be sufficient.

timeline/views_ajax.py
```diff
@@ -211,4 +243,15 @@ def get(self, request, *args, **kwargs):
...
+            if self.kwargs['idx']:
+                status = event.get_status_changes()
+                return Response(
+                    self.get_event_extra(
+                        event, status[int(self.kwargs['idx'])]
+                    ),
+                    status=200,
+                )
+        except KeyError:
+            return Response(self.get_event_extra(event), status=200)
```
mikkonie, line 250: Here we also have quite a bit of repetition between the two views, please refactor.

Also I'm not sure if we only want to return non-status extra data when we encounter a `KeyError`?

timeline/views_ajax.py
```diff
@@ -132,20 +149,24 @@ def _html_print_array(self, array, str_list, indent):
         str_list.append('&nbsp;&nbsp;' * (indent - 1))
         str_list.append('<span class="json-close-bracket">]</span>')
 
-    def get_event_extra(self, event):
+    def get_event_extra(self, event, status=None):
```
mikkonie, line 154: I added extra data to a project event's status update, but for some reason it is showing up empty in the view. Please check:
2023-01_timeline_extra_data_bug

gromdimon: That's weird, by me it's functionating just ok
Screenshot from 2023-01-26 11-25-08
Screenshot from 2023-01-26 11-24-47

timeline/static/timeline/js/timeline.js
gromdimon: Created a helper function

timeline/views_ajax.py
gromdimon: Moved logic to js

timeline/views_ajax.py
gromdimon: I am not sure, that refactoring this logic in a separatet Mixin is a good idea, as mostly this code is retriving GET. However, I optimized it a little bit. Maybe it is ok?

gromdimon:
Also I noticed some layout bugs with status badges in details modal. I have not deeply investigate this problem, because I am not sure in the correctness of implementation. TBD

gromdimon:
Found and fixed layout bug)

timeline/views_ajax.py
mikkonie: Yep, looks better now.

timeline/tests/test_views_ajax.py
```diff
@@ -254,3 +270,88 @@ def test_get(self):
...
+                    kwargs={'projectevent': self.event.sodar_uuid, 'idx': 0},
+                ),
+            )
+        self.assertEqual(response.status_code, 200)
+        expected = {
+            'app': self.event.app,
+            'name': self.event.event_name,
+            'user': self.user.username,
+            'timestamp': self._format_ts(self.event.get_timestamp()),
+            'extra': '<span class="json-open-bracket">{</span><br><span '
```
mikkonie: I don't think we need to assert the whole string of HTML extra data here, just asserting it equals the output of the JSON to HTML generating method is ok :)

timeline/static/timeline/js/timeline.js
```diff
@@ -22,9 +44,16 @@ $(document).ready(function() {
             );
             var tableBody = $('.modal-body').find('tbody');
             for (var i = 0; i < data['status'].length; i++) {
+                let editionalDescription = '<a class="sodar-tl-link-status-extra-data text-primary pull-right" ' +
```
mikkonie: Minor point, but I'm not sure what "editionalDescription" means :) Since this is just an internal variable anyway, just some shorter variable name is sufficient. As long as it's understandable. What this in fact contains is a link to extra data, maybe name it as such?

timeline/tests/test_ui.py
```diff
@@ -296,6 +301,28 @@ def test_extra_modal(self):
...
+        button.click()
+        WebDriverWait(self.selenium, 10).until(
+            EC.presence_of_element_located(
+                (By.CLASS_NAME, 'sodar-tl-link-status-extra-data')
+            )
+        )
+        status_button = self.selenium.find_element(
+            By.CLASS_NAME, 'sodar-tl-link-status-extra-data'
+        )
+        status_button.click()
```
mikkonie: Here we don't assert anything after the last click, so I guess the click can be omitted? If your goal is to assert the element is clickable, you should use `elementToBeClickable` instead.

timeline/views_ajax.py
```diff
@@ -45,6 +45,13 @@ def get_event_details(self, event):
                     'description': s.description,
                     'type': s.status_type,
                     'class': get_status_style(s),
+                    'extra_status_link': '/timeline/ajax/extra/{}/{}'.format(
+                        event.sodar_uuid
+                        if event.project
+                        else f'site/{event.sodar_uuid}',
+                        idx,
+                    ),
+                    'no_extra_status': s.extra_data == {},
```
mikkonie: I think the `no_extra_status` field may be redundant. Just have `extra_status_link` be empty if there is no extra data and check for that instead? Unless there's a case I'm not seeing.

timeline/views_ajax.py
```diff
@@ -45,6 +45,13 @@ def get_event_details(self, event):
                     'description': s.description,
                     'type': s.status_type,
                     'class': get_status_style(s),
+                    'extra_status_link': '/timeline/ajax/extra/{}/{}'.format(
```
mikkonie: On the server side, always use `reverse()` to generate URLs. This way we don't have to go manually changing URLs if the paths get changed (and they occasionally do).

timeline/views_ajax.py
```diff
@@ -132,20 +139,24 @@ def _html_print_array(self, array, str_list, indent):
...
         str_list.append('<span class="json-close-bracket">]</span>')
 
-    def get_event_extra(self, event):
+    def get_event_extra(self, event, status=None):
         """
         Return event extra data.
         :param event: ProjectEvent object
         :return: JSON-serializable dict
         """
+        if status:
```
mikkonie: You could also do this as a one-liner like `extra = status.extra_data if status else event.extra_data` and then call `_json_to_html()` only once when setting up the return value.

timeline/views_ajax.py
mikkonie: I will try to figure out how to reproduce this, I'll get back to you.

timeline/views_ajax.py
gromdimon: The point here is that status extra data can be shown for any status (because with empty extra_data will be formed just {}). 
So to check this in backend side we have to make additional requests. 
To check this in jquery we have to make kind of string matching, what is possible, but in my opinoin not optimal way. 

So now is a following question: what exactly should I implement?

timeline/views_ajax.py
mikkonie: I replicated the event where this fails. When browsed in the UI, the extra data for the status appears empty even though it is `{'test': 'something'}`.

```
from projectroles.plugins import get_backend_api
timeline = get_backend_api('timeline_backend')
User = get_user_model()
u = User.objects.first()
p = Project.objects.filter(type='PROJECT').first()
e = timeline.add_event(project=p, app_name='filesfolders', user=u, event_name='folder_create', description='create folder', extra_data={'test': 'something'})
s = e.set_status('OK', 'All OK', {'test': 'something'})
```

gromdimon: hmm, will see now

timeline/views_ajax.py
mikkonie: I'm not sure if I follow. If the extra data is `{}`, we simply don't show the link. So IMHO it's sufficient to only have the field for the URL and set it empty of the extra data field is empty.

gromdimon: Oh way, I overcomplicated this task in my head) Everything should be alright now

timeline/views_ajax.py
mikkonie: After the most recent updates, I no longer see the link for extra data for the status. But the extra data is there in the object, I even verified to make sure in Django shell. My immediate guess is that something goes wrong when you create the link for the Ajax view.

gromdimon: It's really strange, because it works by me)
Screenshot from 2023-01-30 14-31-49

mikkonie: This works now. The last failure was most likely some browser cache issue on my end..

timeline/tests/test_views_ajax.py
```diff
@@ -254,3 +273,69 @@ def test_get(self):
...
+        )
+        self.event_site_status_init = self.event_site.set_status(
+            'INIT',
+            DEFAULT_MESSAGES['INIT'],
+            extra_data={'test': 'test'},
+        )
+
+    def test_get(self):
+        """Test event status extra data retrieval"""
+        self.maxDiff = None
```
mikkonie: Please clean up this and other manual `maxDiff` settings from tests, they're ok for debugging but let's not commit them :)

mikkonie (changes requested):

Issue #1095, opened 2023-01-13
Add timeline search


mikkonie:
SODAR Core search has not been implemented in timeline and it would be very useful to add.

**NOTE:** This is a rather complex feature, so should be its separate pull request even if other timeline issues are being worked on at the same time.

## Spec

- [See project app search feature development instructions here](https://sodar-core.readthedocs.io/en/dev/dev_project_app.html#project-search-api-and-template)
- Add `search()` implementation to `timeline.plugins.ProjectAppPlugin`
    * Initially we should add support for a single search type: `timeline`. This should search for matches in the `event_name`, `description` and `project__full_title` fields
    * User access to events must be enforced in results, of course
        + For project events, check for `timeline.view_timeline` perm
        + Classified events shown only to users with `timeline.view_classified_event` perm
        + Tests for access must be added
    * Sort by timestamp desc, just like in the timeline views
- Add search template
    * Columns: same ones as in project list
    * Simplified, no modal links for status updates or extra data
- **Hint:** See the filesfolders app for an example of how search is implemented there

## Tasks

- [ ] Add `search()` functionality
- [ ] Add template
- [ ] Test, test, test!

mikkonie:
Done by @gromdimon, closing.

PR #1109, opened 2023-01-26
Timeline search

gromdimon:
Issue #1095

### DIFF
--- a/timeline/models.py
+++ b/timeline/models.py
@@ -60,6 +60,7 @@ class ProjectEventManager(models.Manager):
         for t in search_terms:
             term_query.add(Q(event_name__icontains=t), Q.OR)
             term_query.add(Q(description__icontains=t), Q.OR)
+            term_query.add(Q(event_objects__name__icontains=t), Q.OR)
         return objects.filter(term_query)
 
 
--- a/timeline/models.py
+++ b/timeline/models.py
@@ -60,7 +60,6 @@ class ProjectEventManager(models.Manager):
         for t in search_terms:
             term_query.add(Q(event_name__icontains=t), Q.OR)
             term_query.add(Q(description__icontains=t), Q.OR)
-            term_query.add(Q(project__full_title__icontains=t), Q.OR)
         return objects.filter(term_query)
 
 
--- a/timeline/models.py
+++ b/timeline/models.py
@@ -5,6 +5,7 @@ import uuid
 
 from django.conf import settings
 from django.db import models
+from django.db.models import Q
 
 # Projectroles dependency
 from projectroles.models import Project
@@ -45,6 +46,22 @@ class ProjectEventManager(models.Manager):
             event_objects__object_model=object_model,
             event_objects__object_uuid=object_uuid,
         ).order_by(order_by)
+
+    def find(self, search_terms, keywords=None):
+        """
+        Return events matching the query.
+
+        :param search_terms: Search terms (list of strings)
+        :param keywords: Optional search keywords as key/value pairs (dict)
+        :return: QuerySet of ProjectEvent objects
+        """
+        objects = super().get_queryset().order_by('event_name')
+        term_query = Q()
+        for t in search_terms:
+            term_query.add(Q(event_name__icontains=t), Q.OR)
+            term_query.add(Q(description__icontains=t), Q.OR)
+            term_query.add(Q(project__full_title__icontains=t), Q.OR)
+        return objects.filter(term_query)
 
 
 class ProjectEvent(models.Model):
--- a/timeline/plugins.py
+++ b/timeline/plugins.py
@@ -106,7 +106,7 @@ class ProjectAppPlugin(ProjectAppPluginPoint):
                 key=lambda x: localtime(x.get_timestamp()).strftime(
                     '%Y-%m-%d %H:%M:%S'
                 ),
-                reverse=True,
+                reverse=False,
             )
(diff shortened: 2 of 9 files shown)


gromdimon (commented):
I have one difficulty:
I was trying to test permissions and ui, but for some reason my tests just do not search timeline events.  I've stuck a little bit on that, but hope, that it is not a huge problem.

timeline/plugins.py
```diff
@@ -70,6 +77,58 @@ def get_statistics(self):
...
+                ),
+                reverse=False,
+            )
+        if items:
+            items = [
+                event for event in items if self.check_permission(user, event)
+            ]
+        return {
+            'all': {
+                'title': 'Event names, descriptions and Project Titles',
```
mikkonie: We can rename this to just "Timeline Events".

timeline/templates/timeline/_search_results.html
```diff
@@ -0,0 +1,74 @@
...
+{% if search_results.all.items|length > 0 %}
+  {% include 'projectroles/_search_header.html' with search_title=search_results.all.title result_count=search_results.all.items|length %}
+
+  <table class="table table-striped sodar-card-table sodar-search-table" id="sodar-tl-search-table">
+    <thead>
+      <tr>
+        <th>Timestamp</th>
+        <th>Event</th>
+        <th>Description</th>
+      </tr>
```
mikkonie, line 64: Looks good, a few things I would like to see in this template:
- Add column for `user`
- Add the project badge for events with a project, similar to the site views. Maybe make the project badge an include template so we don't have repetition?

timeline/plugins.py
```diff
@@ -70,6 +77,58 @@ def get_statistics(self):
             }
         }
 
+    def check_permission(self, user, event):
+        """Check if user has permission to view event"""
+        if event.classified:
```
mikkonie: I was able to make the site crash with the timeline search feature, and it seems `check_permission()` is the culprit.

This assumes every event has an associated project, which they of course don't as we can also have site-wide events. So if the search results contain a site-wide event, this will crash with the following:

 ```
AttributeError at /project/search/results/
'NoneType' object has no attribute 'public_guest_access'
```

You should also check for non-project event perms. Please also update tests to include site-wide events to ensure this works.

File: timeline/models.py [hunk 4]
```diff
@@ -46,6 +47,22 @@ def get_object_events(
...
+        :param search_terms: Search terms (list of strings)
+        :param keywords: Optional search keywords as key/value pairs (dict)
+        :return: QuerySet of ProjectEvent objects
+        """
+        objects = super().get_queryset().order_by('event_name')
+        term_query = Q()
+        for t in search_terms:
+            term_query.add(Q(event_name__icontains=t), Q.OR)
+            term_query.add(Q(description__icontains=t), Q.OR)
+            term_query.add(Q(project__full_title__icontains=t), Q.OR)
```
mikkonie: Searching by the project full title was in my original spec. Now, after testing the feature for a bit, I realize I was wrong to design it that way :)

The project title search should be removed, as it just returns a ton of elements for each project matching the substring, which will not be very useful. One can simply browse the project timeline for that.

timeline/templates/timeline/_search_item.html
```diff
@@ -0,0 +1,42 @@
...
+        {{ user_html|safe }}
+      {% else %}
+        N/A
+      {% endif %}
+    </div>
+  </td>
+  {% if event.description %}
+    <td>
+      <div class="sodar-overflow-container">
+        {% if event.project and user.is_superuser %}
```
mikkonie: I was wondering why the badges were not visible when I was testing, but looks like they're only visible for superusers. This condition should be removed, in search results they should be visible for all.

timeline/templates/timeline/_search_results.html
```diff
@@ -0,0 +1,75 @@
...
+  .table#sodar-tl-search-table thead tr th:nth-child(4) {
+    width: 75px !important;
+  }
+
+  .table#sodar-tl-search-table tbody tr td:nth-child(4) {
+    width: 75px !important;
+  }
+
+  /* Responsive modifications */
+  @media screen and (max-width: 1300px) {
```
mikkonie, line 25: These CSS media modifications could use a bit of tweaking for the search result list, but it's a bit hard to describe. I will do it when I clean up the v0.12.0 release :)

timeline/tests/test_models.py
```diff
@@ -269,6 +269,29 @@ def test_repr_no_project_no_user(self):
         expected = "ProjectEvent('N/A', 'test_event', 'N/A')"
         self.assertEqual(repr(self.event), expected)
 
+    def test_find_name(self):
```
mikkonie, line 284: You wrote earlier you had some problems with timeline events and testing. Is that still an issue and if so, can you describe it to me in detail? I can help out.

gromdimon: I've already fixed it! Thanks

mikkonie:
This is looking very good now, all my previous concerns have been addressed. Great work!

One more request though! It seems in my original spec I forgot a very important search parameter. We should also query for `event_objects__name__icontains`.

This brings up results by the event's `ProjectEventObjectRef` references by their display name. It might actually be the most important thing to search here :)

If you can add this and some tests to ensure it's finding what it's supposed to, it would be awesome. At that point we can merge and close this one.

Issue #1112, opened 2023-01-30
Add sodar_uuid field and usage to ProjectEventStatus model


mikkonie:
The `sodar_uuid` field was originally left out from the timeline app `ProjectEventStatus` model, as these models were not often directly accessed. With the ongoing addition of #1096, this is no longer the case. With event status extra data being queried via Ajax from server side, we now have to use an index-based query which is not very SODAR Core-ish.

Instead, we should add the standard UUID field and update the event status extra data Ajax views to use direct references to the `ProjectEventStatus` model.

**NOTE:** Since this feature requires working with migrations, I strongly recommend working on it after all the ongoing PRs have been closed, as moving back and forth between work branches with different migrations can be very painful.

## Spec

- Add `get_project()` method to  `ProjectEventStatus`
    * Should return `event.project`
    * This is needed for SODAR Core project access mixins
- Add `sodar_uuid` field to `ProjectEventStatus`
    * The field needs to be manually populated in migrations, see [here](https://sodar-core.readthedocs.io/en/dev/dev_project_app.html#model-uuid-field) and [here](https://docs.djangoproject.com/en/3.2/howto/writing-migrations/#migrations-that-add-unique-fields) for more information
- Update timeline extra data views and their usage to take a direct `ProjectEventStatus` reference instead of `ProjectEvent`and an index
    * Update link creation for views accordingly, of course
- Update tests as needed

## Tasks

- [ ] Add `get_project()` method
- [ ] Add and populate `sodar_uuid` field
- [ ] Update views and usage
- [ ] Update tests

mikkonie:
Done by @gromdimon.

PR #1115, opened 2023-02-02
status uuid

gromdimon:
Issue #1112 
- [x] Added get_project() method
- [x] Added and populated sodar_uuid field
- [x] Updated views and usage
- [x] Updated tests


timeline/urls.py
```diff
@@ -60,12 +60,12 @@
...
     ),
     url(
-        regex=r'^ajax/extra/(?P<projectevent>[0-9a-f-]+)/(?P<idx>[0-9])$',
+        regex=r'^ajax/extra/(?P<projectevent>[0-9a-f-]+)/(?P<eventstatus>[0-9-a-f-]+)$',
         view=views_ajax.ProjectEventExtraAjaxView.as_view(),
         name='ajax_extra_status_project',
     ),
     url(
-        regex=r'^ajax/extra/site/(?P<projectevent>[0-9a-f-]+)/(?P<idx>[0-9])$',
+        regex=r'^ajax/extra/site/(?P<projectevent>[0-9a-f-]+)/(?P<eventstatus>[0-9-a-f-]+)$',
```
mikkonie: This can be simplified a bit: We don't need the `projectevent`in these URLs, as that can be derived from the `eventstatus` objects.

timeline/views_ajax.py
```diff
@@ -239,12 +228,44 @@ def get(self, request, *args, **kwargs):
...
-                self.get_event_extra(event, status[int(self.kwargs['idx'])]),
-                status=200,
-            )
+
+class EventStatusExtraAjaxView(
+    EventExtraDataMixin, SODARBasePermissionAjaxView
+):
+    """Ajax view for retrieving event status extra data for events"""
+
+    permission_required = 'timeline.view_status_extra'
```
mikkonie: We still use this perm which only has `is_authenticated` as predicate. It should be removed completely.

Also, if we don't check perms within the class at all, only in `get()`, we should use `SODARBaseAjaxView` instead of `SODARBasePermissionAjaxView`. At that point the `permission_required` member can be omitted completely.

timeline/rules.py
```diff
@@ -44,3 +44,6 @@
 
 # Allow viewing classified site-wide event
 rules.add_perm('timeline.view_classified_site_event', rules.is_superuser)
+
+# Allow viewing status extra data
+rules.add_perm('timeline.view_status_extra', rules.is_authenticated)
```
mikkonie: Please remove this, see below.

Issue #321, opened 2019-08-27
Replace ProjectUserTag with project/user specific AppSetting


mikkonie:
Since implemented #266, the `ProjectUserTag` model and related APIs have been useless and should finally be removed.

The model is currently only used to "star" a project in the project header UI. This can instead be done with a `PROJECT_USER` scope app setting, selectable on the UI exactly as now.

The tricky part will be to do the migration scripts: before the migration to remove the `ProjectUserTag` model, we need to have a migration which populates the `AppSetting` model with existing project starring setting values.

We should also document this in major changes as it is theoretically possible that some site using SODAR Core would have used this model for something else. I'm 99.9998% sure this hasn't happened though :)

## Spec

- First get to know app settings, if not familiar yet
    * [AppSettingAPI](https://sodar-core.readthedocs.io/en/dev/app_projectroles_api_django.html?highlight=app%20settings#app-settings)
    * [Projectroles app settings](https://sodar-core.readthedocs.io/en/dev/dev_core_resource.html?highlight=app%20settings#projectroles-app-settings)
- Define a new setting in `PROJECTROLES_APP_SETTINGS`
    * `name = project_star`
    * `scope = APP_SETTING_SCOPE_PROJECT_USER`
    * `default = False`
- Add new `projectroles` migration to populate settings according to existing `ProjectUserTag` objects
    * For each object, create a setting for the project & user using `AppSettingAPI`
    * For reverse code, delete all app settings where `app == projectroles`  and `name == project_star`
- Update `ProjectStarringAjaxView` to use app settings instead of `ProjectUserTag`
- Update usages of `get_tag_state()` accordingly
- Update view and UI tests as needed to fit this new functionality
- **Only after testing the aforementioned steps properly:**
    * Remove `projectroles.project_tags`
    * Remove `ProjectUserTag` model and make migrations
    * Update docs

gromdimon:
### TODO:
- [ ] Change home template to show propper stars

mikkonie:
Done by @gromdimon.

PR #1126, opened 2023-02-09
Replace ProjectUserTag with app setting

gromdimon:
Issue # 321


gromdimon:
Issue #321

projectroles/app_settings.py
```diff
@@ -79,6 +79,15 @@
         'user_modifiable': True,
         'local': False,
     },
+    'project_star': {
+        'scope': APP_SETTING_SCOPE_PROJECT_USER,
+        'type': 'BOOLEAN',
+        'default': False,
+        'label': 'Project Star',
```
mikkonie: `label`, `description` and `user_modifiable` are not used with app settings of this scope, so they should be removed. This may not be properly explained in the SODAR Core docs, I will improve that.

projectroles/migrations/0024_project_star_app_setting.py
```diff
@@ -0,0 +1,32 @@
+from django.db import migrations
+
+
+def migrate_stars(apps, schema_editor):
+    """Set rank values for existing roles"""
```
mikkonie: This docstring could use some updating :)

projectroles/migrations/0024_project_star_app_setting.py
```diff
@@ -0,0 +1,32 @@
...
+    ProjectUserTag = apps.get_model('projectroles', 'ProjectUserTag')
+    Project = apps.get_model('projectroles', 'Project')
+    AppSetting = apps.get_model('projectroles', 'AppSetting')
+    for project in Project.objects.all():
```
mikkonie: You can just query for all `ProjectUserTag` objects, no need to have nested queries per project. Just in case, filter for `name=PROJECT_TAG_STARRED` in case some site would have used this model for things other than the starring functionality.

projectroles/migrations/0024_project_star_app_setting.py
```diff
@@ -0,0 +1,32 @@
...
+        user_tag = ProjectUserTag.objects.filter(project=project).first()
+        if user_tag is not None:
+            setting = AppSetting()
```
mikkonie: As said previously, please use `get_or_create()` here. The fields can be passed on as kwargs for that method.

projectroles/migrations/0024_project_star_app_setting.py
```diff
@@ -0,0 +1,32 @@
...
+            setting.project = project
+            setting.user = user_tag.user
+            setting.value = True
+            setting.name = 'project_star'
+            setting.user_modifiable = True
```
mikkonie: `user_modifiable` is not relevant for the scope of this setting, so we don't need to define it here and it can simply remain default.

projectroles/project_tags.py
```diff
@@ -1,59 +1,24 @@
 """Functions for project tagging/starring in the projectroles app"""
 # NOTE: This can be expanded to include other types of tags later on
 
-from projectroles.models import ProjectUserTag, PROJECT_TAG_STARRED
+from projectroles.app_settings import AppSettingAPI
```
mikkonie: The whole `project_tags` module should be removed, as we will no longer have the concept of a project tag. Project starring/unstarring should happen within the relevant Ajax view.

projectroles/project_tags.py
```diff
@@ -1,59 +1,24 @@
...
-    :param name: Tag name (string)
-    :return: Boolean
-    """
-    if not user.is_authenticated:
-        return False
 
-    try:
-        ProjectUserTag.objects.get(project=project, user=user, name=name)
-        return True
+def get_tag_state(project, user):
```
mikkonie: Also, we can just call `app_settings.set()` and `get()`, no need for further helpers I think.

projectroles/views.py
```diff
@@ -1648,7 +1647,7 @@ def delete_assignment(self, request, instance):
         # Delete object itself
         instance.delete()
         # Remove project star from user if it exists
-        remove_tag(project=project, user=user)
+        set_tag_state(project, user, star=False)
```
mikkonie: We should not set this setting to false here. Instead, we should delete it using `app_settings.delete()`.

In fact, we should probably delete all `PROJECT-USER` scope settings for the project & user combination when a user's role is removed from a project. But that's a separate issue, I'll think about it and open a ticket..

projectroles/tests/test_views.py
```diff
@@ -1968,6 +1968,16 @@ def setUp(self):
             project=self.project,
         )
 
+        # Init project_star setting
```
gromdimon: To be honest I do not understan, why this get/post view tests are working when I add this setting and do not without it. The probem went deeply to AppSettingAPI methods, where I was little bit lost. So it is working as it is, but I do not know why(

projectroles/migrations/0024_project_star_app_setting.py
```diff
@@ -0,0 +1,29 @@
+from django.db import migrations
+
+
+def migrate_stars(apps, schema_editor):
+    """Create project_star AppSettings"""
+    ProjectUserTag = apps.get_model('projectroles', 'ProjectUserTag')
+    AppSetting = apps.get_model('projectroles', 'AppSetting')
+    for projectusertag in ProjectUserTag.objects.all():
```
mikkonie: Just a minor note, it is convention to name variables with underscores between words. Also since this is just a local control structure, you can use something simple like `tag`.

Also, as I requested before: filter objects by `name=PROJECT_TAG_STARRED`, just in case.

projectroles/tests/test_views.py
mikkonie: Can you clarify on what exactly was not working and how?

projectroles/templatetags/projectroles_tags.py
```diff
@@ -68,7 +68,7 @@ def has_star(project, user):
     """Return True/False for project star tag state"""
     return user.has_perm(
         'projectroles.view_project', project
-    ) and get_tag_state(project, user, PROJECT_TAG_STARRED)
```
mikkonie, line 71: The original implementation of this looked a bit weird to me, so I had a closer look. Turns out the `has_star()` template tag is no longer used for anything nowadays. I think this is residue from a time when the site had a server side project list.

This template tag and tests related to it can be removed.

projectroles/tests/test_templatetags.py
```diff
@@ -453,15 +451,27 @@ def test_get_backend_plugins(self):
 
     # TODO: Test get_site_app_messages() (set up admin alert)
 
-    def test_has_star(self):
-        """Test has_star()"""
-        # Test with no star
+    def test_star(self):
```
mikkonie: As said above, these tests can be removed together with the template tag itself.

projectroles/tests/test_views_ajax.py
```diff
@@ -343,48 +345,35 @@ def setUp(self):
             self.project, self.user, self.role_owner
         )
 
-    def test_star_project(self):
-        """Test project starring"""
-        self.assertEqual(ProjectUserTag.objects.all().count(), 0)
-
+    def test_project_star(self):
+        """Test Starring a Project"""
+        # Test star
```
mikkonie: This comment seems a bit redundant and can be removed.

projectroles/views_ajax.py
```diff
@@ -368,9 +368,21 @@ def post(self, request, *args, **kwargs):
         project = self.get_project()
         user = request.user
         timeline = get_backend_api('timeline_backend')
-        tag_state = get_tag_state(project, user)
+        tag_state = app_settings.get(
```
mikkonie: The concept of "tag" will no longer exist after this, so the variable should also be renamed. Otherwise it might be a bit confusing :)

projectroles/tests/test_views.py
gromdimon: If I delete this project_star setting in test, than I become failure in requests:
```response = self.client.get(
              reverse(
                  'projectroles:update',
                  kwargs={'project': self.project.sodar_uuid},
              )
          )
```
The exception is being made in `app_settings.py` file in:
```raise KeyError(
            'Setting "{}" not found in app plugin "{}"'.format(
                setting_name, app_name
            )
        )
```
I am just confused, is it looking for every app_setting for projectroles?

PR #1132, opened 2023-02-13
Timeline sorting features

### LINKED ISSUE #1110: Timeline sorting/filtering
### Problem

Once, I noticed, that additional sorting in timeline would a unseful feature. So why not to try something?
[...]

### Solution

I found a good example in projectroles sorrting/filtering search results logic. Basically it is pure ajax with asynchronous template manipulation. 

### Implementation
- [ ] Add `sort.js` ajax
- [ ] Update timeline templates
- [ ] Test it!

### Alternative Solutions

As alternative I see a special function with parameters for sort (name, timestamp, etc.), that just make filtering in backend and re-sends results back to template. However, I think it'll be much slower because of sending, retriving and re-rendering requests.

### Additional Context

I've decided to start work on the problem just because I was gone out of tasks. Hope, it is usefull and right.

gromdimon:
Init

### DIFF
--- a/timeline/static/timeline/css/timeline.css
+++ b/timeline/static/timeline/css/timeline.css
@@ -24,6 +24,63 @@ table#sodar-tl-table-extra-data tbody tr td:nth-child(1) {
   width: 100%;
 }
 
+table.sodar-timeline-table {
+  display: inline-table;
+  min-width: 100%;
+  max-width: 100%;
+  width: 100%;
+  border-bottom: 0 !important;
+  margin: 0 !important;
+}
+
+table.sodar-timeline-table thead tr th {
+  padding: 12px !important;
+  padding-right: 22px !important; /* Avoid covering sort icons */
+  max-width: 250px !important;
+}
+
+table.sodar-timeline-table tbody tr td {
+  padding: 12px !important;
+  max-width: 250px !important;
+}
+
+.sodar-header-input-group-search {
+    max-width: 250px;
+}
+
+@media screen and (max-width: 1000px) {
+  .sodar-header-input-group-search {
+    display: none;
+  }
+}
+
+select.sodar-sort-page-length {
+    height: 32px;
+    padding-top: 5px;
+}
+
+th.sorting {
+  padding-left: 10px !important;
+  border-bottom: 0 !important;
+}
+
+th.sorting::after {
+  padding-top: 7px !important;
+}
+
+.dataTables_paginate {
+  width: 100%;
+  padding: 12px !important;
+  margin-bottom: 12px !important;
+  text-align: right;
+  border-top: 1px solid #dfdfdf !important;
+}
+
+.dataTables_paginate .pagination .active .page-link {
+    background-color: #007bff;
+    color: #ffffff;
+}
+
 a.sodar-tl-link-detail:hover {
   text-decoration: underline;
 }
--- a/timeline/static/timeline/css/timeline.css
+++ b/timeline/static/timeline/css/timeline.css
@@ -24,7 +24,7 @@ table#sodar-tl-table-extra-data tbody tr td:nth-child(1) {
   width: 100%;
 }
 
-table.sodar-timeline-table {
+table.sodar-tl-table {
   display: inline-table;
   min-width: 100%;
   max-width: 100%;
@@ -33,25 +33,15 @@ table.sodar-timeline-table {
   margin: 0 !important;
 }
 
-table.sodar-timeline-table thead tr th {
+table.sodar-tl-table thead tr th {
   padding: 12px !important;
   padding-right: 22px !important; /* Avoid covering sort icons */
   max-width: 250px !important;
 }
 
-table.sodar-timeline-table tbody tr td {
+table.sodar-tl-table tbody tr td {
   padding: 12px !important;
   max-width: 250px !important;
-}
-
-.sodar-header-input-group-search {
-    max-width: 250px;
-}
-
-@media screen and (max-width: 1000px) {
-  .sodar-header-input-group-search {
-    display: none;
-  }
 }
 
 select.sodar-sort-page-length {
--- a/timeline/templates/timeline/_list_item.html
+++ b/timeline/templates/timeline/_list_item.html
@@ -4,7 +4,6 @@
 
 <tr id="sodar-tl-list-event-{{ event.sodar_uuid }}">
   <td>
-    <div class="sodar-overflow-container">
       {% get_timestamp event as event_time %}
       {% if event.project %}
         <a class="sodar-tl-link-detail text-primary"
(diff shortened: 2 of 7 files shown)


gromdimon:
Issue https://github.com/bihealth/sodar-core/issues/1110

- [x] Implemented losorting logic for project, site and admin timelines
- [x] Added tests (However most of the logic is on jquery and we do not test it, that is why "real" tests are ommited)

Looking forward for any comments!

timeline/static/timeline/css/timeline.css
```diff
@@ -24,6 +24,63 @@ table#sodar-tl-table-extra-data tbody tr td:nth-child(1) {
   width: 100%;
 }
 
+table.sodar-timeline-table {
```
mikkonie: It is convention to name SODAR Core app-specific CSS classes using `sodar-` followed by a two-letter abbreviation of the app. That would make this `sodar-tl-table` (or something else if that's already taken).

timeline/static/timeline/css/timeline.css
```diff
@@ -24,6 +24,63 @@ table#sodar-tl-table-extra-data tbody tr td:nth-child(1) {
...
+  padding-right: 22px !important; /* Avoid covering sort icons */
+  max-width: 250px !important;
+}
+
+table.sodar-timeline-table tbody tr td {
+  padding: 12px !important;
+  max-width: 250px !important;
+}
+
+.sodar-header-input-group-search {
```
mikkonie: Same here, if this is timeline specific please prefix with `sodar-tl-*`. If it's meant to be a commonly used class, add in `projectroles.css` instead.

timeline/templates/timeline/_list_item.html
```diff
@@ -4,56 +4,73 @@
...
-         data-url="{% url 'timeline:ajax_detail_project' projectevent=event.sodar_uuid %}">
-        {{ event_time }}
-      </a>
-    {% else %}
-      <a class="sodar-tl-link-detail text-primary"
-         data-url="{% url 'timeline:ajax_detail_site' projectevent=event.sodar_uuid %}">
-        {{ event_time }}
-      </a>
-    {% endif %}
+    <div class="sodar-overflow-container">
```
mikkonie: The way the table CSS is set, I don't believe we need to use `sodar-overflow-container` in any of these cells. The content of any column should not overflow in a way we would like to expand with mouseover. The description column content can split on multiple rows and the others are basically fixed. Unless there's something I'm missing.

timeline/templates/timeline/_sort_header.html
```diff
@@ -0,0 +1,21 @@
+<div class="card sodar-sort-card">
+  <div class="card-header">
+    <h4>
+      <i class="iconify"
+         data-icon="mdi:sort-clock-ascending"></i>
+      Sort options
```
mikkonie: "Timeline Events" would be an appropriate title for the card. Also note that [title case](https://research.arizona.edu/faq/what-do-you-mean-when-you-say-use-title-case-proposalproject-titles) should be used for all titles. I probably haven't specified this in contibuting docs, will do so when I do cleanup for the next release :)

timeline/templates/timeline/_sort_header.html
```diff
@@ -0,0 +1,21 @@
+<div class="card sodar-sort-card">
+  <div class="card-header">
+    <h4>
+      <i class="iconify"
```
mikkonie: We can omit the icon here, it's not necessary.

timeline/templates/timeline/_sort_header.html
```diff
@@ -0,0 +1,21 @@
+<div class="card sodar-sort-card">
```
mikkonie: There's a slight bug here: we are including a card within a card, resulting in two cards being displayed on top of each other in the layout.

timeline/templates/timeline/timeline.html
```diff
@@ -65,9 +71,22 @@ <h3><i class="iconify" data-icon="mdi:clock-time-eight"></i> {{ timeline_title }
 {% block javascript %}
   {{ block.super }}
 
+  <!-- Settings for Javascript -->
+  <script type="text/javascript">
```
mikkonie: This section is for the search results page and should not be included for the timeline view.

timeline/templates/timeline/timeline.html
```diff
@@ -36,17 +40,19 @@ <h3><i class="iconify" data-icon="mdi:clock-time-eight"></i> {{ timeline_title }
...
+            <tbody>
+              {% for event in object_list %}
+                {% include 'timeline/_list_item.html' %}
+              {% endfor %}
+            </tbody>
+          </table>
+        </div>
       </div>
     </div>
     {% if is_paginated %}
```
mikkonie: We don't need to include the server-side pagination template anymore if we're doing it on the client side.

mikkonie:
The UI works really well and the features are desireable. However, in addition to my review comments I see one major issue regarding this implementation: the actual view where we query for events is unchanged.

This means that:

1. As long as we still use server-side pagination, we will only get the maximum amount of results for the first server-side "page" of results and are unable to access the rest.
2. If we remove the server-side pagination, this means we will query for all relevant timeline events at once and then handle them in the UI. This will not scale well, especially on the "all timeline events" admin view, but also not for longstanding projects wtih a lot of events.

You can give this a try yourself. Make a loop in django cell which generates several thousand dummy events (preferably with object references, as that requires extra queries and is realistic for actual events) and see how it performs after removing the server-side pagination. For reference, the SODAR production server currently has ~58000 events :)

For this to work in a scalable way, I believe we need to add a new Ajax view for event retrieval, which the DataTables can then query. I'm guessing this should be doable with DataTables, but never done it myself. Feel free to investigate.

timeline/templates/timeline/_sort_header.html
```diff
@@ -0,0 +1,21 @@
...
+  <div class="card-header">
+    <h4>
+      <i class="iconify"
+         data-icon="mdi:sort-clock-ascending"></i>
+      Sort options
+      <div class="input-group sodar-header-input-group sodar-header-input-group-sort pull-right">
+        <select class="form-control sodar-sort-page-length">
```
mikkonie: Forgot to comment this in my review: this select box breaks the layout as it is too high. Adding a `sodar-header-input-group select` definition in `projectroles.css` with the appropriate height should do the trick.

timeline/static/timeline/css/timeline.css
gromdimon: my fault, renamed

timeline/static/timeline/css/timeline.css
gromdimon: It was already in `projectroles.css`)

timeline/templates/timeline/_list_item.html
gromdimon: yep, deleted

timeline/templates/timeline/_sort_header.html
gromdimon: Moved sodar-sort-card class into the first card div

timeline/templates/timeline/timeline.html
gromdimon: But this section corresponds to pagination on client site. As it was mentioned in discussion there should be some way of making pagination on server side and then sorting options on client one. So I will change it

gromdimon:
> 

About pagination issue:
I just removed client-side pagination and now dataTable filters results on the page but pagination is handled server-side.

mikkonie:
I've converted this into draft since it is not relevant at this time. Should we close the entire PR for now instead?

PR #1134, opened 2023-02-15
Prevent sending invites to local users if local users are disabled

### LINKED ISSUE #616: Prevent sending invites to local users if local users are disabled
If local users are not allowed, we should also not allow sending invites to users not recognized in our LDAP setup.

## Spec

- Logic for denial
    * IF `PROJECTROLES_ALLOW_LOCAL_USERS==False`
    * AND `ENABLE_SAML==False`
    * AND email domain not in `AUTH_LDAP_USERNAME_DOMAIN`, `AUTH_LDAP2_USERNAME_DOMAIN` or `ALT_LDAP_DOMAINS`
    * THEN prevent creation of invite
- Where and how to implement this
    * Add check in `ProjectInviteForm.clean()`
    * Perform check, use `add_error()` on the `email`field
    * Error message should be *"Local users not allowed, email domain not recognized for LDAP users"*
- Add test(s) in `TestProjectInviteCreateView`

gromdimon:
Issue #616 
Done:
- [x] Added check in ProjectInviteForm.clean()
- [x] Added tests in TestProjectInviteCreateView

### DIFF
--- a/projectroles/forms.py
+++ b/projectroles/forms.py
@@ -962,7 +962,7 @@ class ProjectInviteForm(SODARModelForm):
         domain = user_email[user_email.find('@') + 1 :]
         domain_list = [
             getattr(settings, 'AUTH_LDAP_USERNAME_DOMAIN'),
-            getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN')
+            getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN'),
         ]
         if (
             not settings.PROJECTROLES_ALLOW_LOCAL_USERS
--- a/projectroles/forms.py
+++ b/projectroles/forms.py
@@ -958,22 +958,28 @@ class ProjectInviteForm(SODARModelForm):
             pass
 
         # Local users check
-        user_email = self.cleaned_data.get('email')
-        domain = user_email[user_email.find('@') + 1 :]
-        domain_list = [
-            getattr(settings, 'AUTH_LDAP_USERNAME_DOMAIN', ''),
-            getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN', ''),
-        ]
-        if (
-            not settings.PROJECTROLES_ALLOW_LOCAL_USERS
-            and not settings.ENABLE_SAML
-            and domain not in domain_list
-        ):
-            self.add_error(
-                'email',
-                'Local users not allowed, email domain {} not recognized for '
-                'LDAP users'.format(domain),
+        try:
+            user_email = self.cleaned_data.get('email')
+            domain = (
+                user_email[user_email.find('@') + 1 :].split('.')[0].lower()
             )
+            domain_list = [
+                getattr(settings, 'ALT_LDAP_DOMAINS', ''),
+                getattr(settings, 'AUTH_LDAP_USERNAME_DOMAIN', ''),
+                getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN', ''),
+            ]
+            if (
+                not settings.PROJECTROLES_ALLOW_LOCAL_USERS
+                and not settings.ENABLE_SAML
+                and domain not in [d.lower() for d in domain_list]
+            ):
+                self.add_error(
+                    'email',
+                    'Local users not allowed, email domain {} not recognized for '
+                    'LDAP users'.format(domain),
+                )
+        except AttributeError:
+            pass
 
         # Delegate checks
         role = self.cleaned_data.get('role')
--- a/projectroles/forms.py
+++ b/projectroles/forms.py
@@ -961,8 +961,8 @@ class ProjectInviteForm(SODARModelForm):
         user_email = self.cleaned_data.get('email')
         domain = user_email[user_email.find('@') + 1 :]
         domain_list = [
-            getattr(settings, 'AUTH_LDAP_USERNAME_DOMAIN'),
-            getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN'),
+            getattr(settings, 'AUTH_LDAP_USERNAME_DOMAIN', None),
+            getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN', None),
         ]
         if (
             not settings.PROJECTROLES_ALLOW_LOCAL_USERS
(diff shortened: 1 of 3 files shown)


projectroles/forms.py
```diff
@@ -957,6 +957,28 @@ def clean(self):
         except ProjectInvite.DoesNotExist:
             pass
 
+        # Local users check
+        user_email = self.cleaned_data.get('email')
+        domain = user_email[user_email.find('@') + 1 :]
+        allow_local_users = getattr(
+            settings, 'PROJECTROLES_ALLOW_LOCAL_USERS', False
```
mikkonie: A very minor thing: Both `PROJECTROLES_ALLOW_LOCAL_USERS` and `ENABLE_SAML` are mandatory settings, so we can refer to them directly via `settings.*` and simplify this a bit.

Other than that, everything looks great!

gromdimon: I tried to change it, but then tests failed( I think for overriding settings the variable should be set with getattr(). Should I change the tests somehow or let it be as is?

projectroles/forms.py
```diff
@@ -957,6 +957,24 @@ def clean(self):
...
+        domain = user_email[user_email.find('@') + 1 :]
+        domain_list = [
+            getattr(settings, 'AUTH_LDAP_USERNAME_DOMAIN', ''),
+            getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN', ''),
+        ]
+        if (
+            not settings.PROJECTROLES_ALLOW_LOCAL_USERS
+            and not settings.ENABLE_SAML
+            and domain not in domain_list
+        ):
```
mikkonie: We are missing the check for `LDAP_ALT_DOMAINS` here. That also must be included as stated in the original spec.

**Note:** While `AUTH_LDAP*_USERNAME_DOMAIN` is in the form of e.g. `CHARITE` or `MDC-BERLIN` (without the top level domain), `LDAP_ALT_DOMAINS` is a list with TLDs, e.g. `['bih-charite.de']`.

Also, use `.lower()` on each value before comparison, as `AUTH_LDAP*_USERNAME_DOMAIN` values are expected as upper case.

projectroles/forms.py
```diff
@@ -957,6 +957,24 @@ def clean(self):
         except ProjectInvite.DoesNotExist:
             pass
 
+        # Local users check
+        user_email = self.cleaned_data.get('email')
+        domain = user_email[user_email.find('@') + 1 :]
```
mikkonie: When comparing against `AUTH_LDAP*_USERNAME_DOMAIN`, note that the expected values here don't include the top level domain. Hence we should be comparing them to the email domain without the TLD. See more in the other comment.

projectroles/tests/test_views.py
```diff
@@ -3123,6 +3123,81 @@ def test_render_not_found(self):
             )
         self.assertEqual(response.status_code, 404)
 
+    @override_settings(
+        PROJECTROLES_ALLOW_LOCAL_USERS=False,
+        ENABLE_SAML=False,
+        AUTH_LDAP_USERNAME_DOMAIN='',
```
mikkonie: Note that this settings value is never set or read unless `ENABLE_LDAP` is true.

projectroles/tests/test_views.py
```diff
@@ -3123,6 +3123,81 @@ def test_render_not_found(self):
...
+        self.assertEqual(response.status_code, 302)
+        invite = ProjectInvite.objects.get(
+            project=self.project, email=INVITE_EMAIL, active=True
+        )
+        self.assertIsNotNone(invite)
+
+    @override_settings(
+        PROJECTROLES_ALLOW_LOCAL_USERS=False,
+        ENABLE_SAML=False,
+        AUTH_LDAP_USERNAME_DOMAIN='example.com',
```
mikkonie: `AUTH_LDAP*_USERNAME_DOMAIN` should not include the top level domain and it is expected in upper case, e.g. `CHARITE`. Not sure if this is properly documented, if not then sorry.

projectroles/forms.py
```diff
@@ -957,6 +957,30 @@ def clean(self):
...
             pass
 
+        # Local users check
+        try:
+            user_email = self.cleaned_data.get('email')
+            domain = (
+                user_email[user_email.find('@') + 1 :].split('.')[0].lower()
+            )
+            domain_list = [
+                getattr(settings, 'ALT_LDAP_DOMAINS', ''),
```
mikkonie: This should be `LDAP_ALT_DOMAINS`. Also note that, as I described in my comment before, the syntax for `LDAP_ALT_DOMAINS` is different. There we have the entire domain with the TLD (e.g. "example.com") while in the primary auth domains we have it without the TLD (e.g. "EXAMPLE").

projectroles/tests/test_views.py
```diff
@@ -3123,6 +3123,80 @@ def test_render_not_found(self):
...
+        invite = ProjectInvite.objects.get(
+            project=self.project, email=INVITE_EMAIL, active=True
+        )
+        self.assertIsNotNone(invite)
+
+    @override_settings(
+        PROJECTROLES_ALLOW_LOCAL_USERS=False,
+        ENABLE_SAML=False,
+        ENABLE_LDAP=True,
+        AUTH_LDAP_USERNAME_DOMAIN='EXAMPLE',
```
mikkonie, line 3177: Also, please add test using `LDAP_ALT_DOMAINS`.

projectroles/forms.py
```diff
@@ -957,6 +957,29 @@ def clean(self):
...
+                user_email[user_email.find('@') + 1 :].split('.')[0].lower()
+            )
+            domain_list = getattr(settings, 'LDAP_ALT_DOMAINS', []) + [
+                getattr(settings, 'AUTH_LDAP_USERNAME_DOMAIN', ''),
+                getattr(settings, 'AUTH_LDAP2_USERNAME_DOMAIN', ''),
+            ]
+            if (
+                not settings.PROJECTROLES_ALLOW_LOCAL_USERS
+                and not settings.ENABLE_SAML
+                and not any(domain in d.lower() for d in domain_list)
```
mikkonie: As stated before, we need to compare the values in `LDAP_ALT_DOMAINS` to the entire domain of the recipient email containing the TLD. Right now we split the domain to not include the TLD and use that for all comparisons. This is going to need two conditions in the if statement for `AUTH_LDAP*_USERNAME_DOMAIN` (without TLD) and `LDAP_ALT_DOMAINS` (with TLD).

PR #1135, opened 2023-02-15
Search settings

### LINKED ISSUE #1119: Add Django setting to omit apps from search results
After we added timeline search in #1095, I started thinking that not all sites necessarily want to enable search on all apps.

For example, timeline information might not be useful for all sites. Also, we have one site which uses the filesfolders app in the background hidden from users, so the search results probably won't be too relevant either.

There should be a way to omit desired apps from search results by the site's Django settings.

## Spec

- Add Django setting `PROJECTROLES_SEARCH_OMIT_APPS` to `base.py`
    * Should be of type list, default being `[]`
    * Preferably insert setting below `PROJECTROLES_ENABLE_SEARCH`
- In `ProjectSearchResultsView._get_app_results()`, omit apps from `search_apps` if on list
    * **NOTE:** The setting should be **optional**: access with `getattr()` in code, avoid Django crash if not set on site
- Add tests to ensure omission works
- Document in `app_projectroles_settings.rst` under "Optional Projectroles Settings"

### LINKED ISSUE #1124: Add item limit to timeline search
Experimenting with timeline search on the SODAR production server, I noticed that one can come up with a search which takes up a very long time and returns a lot (in this case ~7000) results. This will take a long time and will most likely not yield much useful results.

Implementing #22 will eventually fix such issues, but that's not in the priorities right now. And still, this search should be limited a bit for meaningful results.

Hence I propose we do a similar thing to iRODS file search limitation in SODAR:

## Spec

- Add `TIMELINE_SEARCH_LIMIT` Django setting
    * Integer, retrieve from env, default=250
- Use limit in timeline search
- Add variable to `info_settings` in `timeline.plugins.ProjectAppPlugin`

### LINKED ISSUE #1125: Cleanup repetition in timeline search()
I just noticed this, I missed it in my earlier review so my bad :)

```
        if not search_type:
            events = ProjectEvent.objects.find(search_terms, keywords)
            items = list(events)
            items.sort(
                key=lambda x: localtime(x.get_timestamp()).strftime(
                    '%Y-%m-%d %H:%M:%S'
                ),
                reverse=False,
            )
        elif search_type == 'timeline':
            events = ProjectEvent.objects.find(search_terms, keywords)
            items = list(events)
            items.sort(
                key=lambda x: localtime(x.get_timestamp()).strftime(
                    '%Y-%m-%d %H:%M:%S'
                ),
                reverse=False,
            )
```

This can be simplified as `if not search_type or search_type == 'timeline':`

Low priority, can be done e.g. together with #1124.

gromdimon:
Issue #1124 and #1119


gromdimon:
Done:
- [x] Added TIMELINE_SEARCH_LIMIT Django setting
- [x] Used limit in timeline search
- [x] Added variable to info_settings in `timeline.plugins.ProjectAppPlugin`
- [x] Added tests
- [x] Added Django setting `PROJECTROLES_SEARCH_OMIT_APPS` to `base.py `
- [x] Implemented logic
- [x] Added tests

timeline/plugins.py
```diff
@@ -131,7 +125,7 @@ def search(self, search_terms, user, search_type=None, keywords=None):
             'all': {
                 'title': 'Timeline Events',
                 'search_types': ['timeline'],
-                'items': items,
+                'items': items[:search_limit],
```
mikkonie: Here, I would recommend instead using the limit already in `ProjectEventManager.find()`. In this case, we query for the maximum amount of objects, convert the queryset into a list, sort that and only then apply the filter. Django querysets are lazy, so if we apply `[:search_limit]` already to `ProjectEvent.objects.filter()`, it will perform much better with large results.

This requires also moving the sorting of events timestamp from here into `find()`. So the query becomes something like `ProjectEvent.objects.filter(etc etc).order_by('-timestamp')[:search_limit]`.

projectroles/views.py
```diff
@@ -588,19 +588,28 @@ def _get_app_results(self, search_terms, search_type, search_keywords):
...
 
         if search_type:
             search_apps = sorted(
                 [
                     p
                     for p in plugins
-                    if (p.search_enable and search_type in p.search_types)
+                    if (
+                        p.search_enable
+                        and search_type in p.search_types
```
mikkonie: The code could be shortened like follows:
```
search_apps = sorted([p for p in plugins if (p.search_enable and p.name not in omit_apps_list)], key= lambda x: x.plugin_ordering)
if search_type:
    search_apps = [p for plugin in search_apps if search_type in p.search_types]
```

timeline/models.py
```diff
@@ -55,13 +55,15 @@ def find(self, search_terms, keywords=None):
...
-        objects = super().get_queryset().order_by('event_name')
+        search_limit = getattr(settings, 'TIMELINE_SEARCH_LIMIT', 250)
+        objects = super().get_queryset().order_by('-status_changes__timestamp')
         term_query = Q()
         for t in search_terms:
             term_query.add(Q(event_name__icontains=t), Q.OR)
             term_query.add(Q(description__icontains=t), Q.OR)
             term_query.add(Q(event_objects__name__icontains=t), Q.OR)
-        return objects.filter(term_query)
+        items = objects.filter(term_query)[:search_limit]
```
mikkonie: We could simplify this into `super().get_queryset().filter(term_query).order_by('-status_changes__timestamp')`.

timeline/tests/test_ui.py
```diff
@@ -423,43 +424,71 @@ def setUp(self):
             app_name='projectroles',
             user=self.superuser,
             event_name='test_event',
-            description='description',
+            description='KsdfLAfANd',
```
mikkonie: Maybe using a local constant here would be the best :)

mikkonie:
Something I forgot to add in the original spec: Please also document the `TIMELINE_SEARCH_LIMIT` django setting in `app_timeline_install.rst` doc, under "Optional Settings". And when editing the docs, don't forget to ensure building the docs works without errors before committing.

mikkonie:
One thing I forgot: please document `PROJECTROLES_SEARCH_OMIT_APPS` in `app_projectroles_settings.rst`, under "Optional Projectroles Settings".

Other than that everything looks very good, will merge once the docs are updated.

PR #1138, opened 2023-02-20
Change regex in urls

### LINKED ISSUE #1116: Update URL patterns to use path instead of regex where applicable
Since implementing #1047, there should be no problem in replacing the URL patterns for `sodar_uuid` URLs with a path syntax instead of the traditional regex. This should make the URL definitions look a lot more readable.

**NOTE:** This is a low priority issue, good to do when e.g. waiting for feedback for some larger task, or if I'm slow to assign stuff and you run out of other tasks :)

## Spec

- In all apps, update `regex` URLs to use `route`
  * As an example, `regex=r'^update/(?P<project>[0-9a-f-]+)$'` would become `route='update/<uuid:project>'`.
  * Preserving the original kwarg names is critical, as those are used by `ProjectAccessMixin` to determine perms.
  * There might be some 
- **Exception:** Leave URLs in `example_project_app` as is, because in there we ensure the working of both regex and path URLs
- Running existing tests should catch any problems with forming URLs in case there's a mistake or typo.
  * No new tests should be required.
- Docs and examples within should be updated to reflect that the recommended way of setting URLs is using `route`, but `regex` also works.
  * See [dev_project_app.rst](https://sodar-core.readthedocs.io/en/dev/dev_project_app.html#url-keyword-arguments)

## Tasks

- [ ] Update URL patterns
- [ ] Ensure tests run
- [ ] Update docs as needed

gromdimon:
Change regex in urls

### DIFF
--- a/adminalerts/urls.py
+++ b/adminalerts/urls.py
@@ -9,22 +9,22 @@ app_name = 'adminalerts'
 
 urls_ui = [
     path(
-        route='list/',
+        route='list',
         view=views.AdminAlertListView.as_view(),
         name='list',
     ),
     path(
-        route='detail/<uuid:adminalert>/',
+        route='detail/<uuid:adminalert>',
         view=views.AdminAlertDetailView.as_view(),
         name='detail',
     ),
     path(
-        route='create/',
+        route='create',
         view=views.AdminAlertCreateView.as_view(),
         name='create',
     ),
     path(
-        route='update/<uuid:adminalert>/',
+        route='update/<uuid:adminalert>',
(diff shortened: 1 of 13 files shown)


gromdimon:
Issue #1116 
Done:
- [x] Update URL patterns
- [x] Ensure tests run

gromdimon:
I found a strange thing, while edditing urls (and handling test errors):
In tokens there is `AuthToken` object and it has pk with letters and numbers (when it's usually just int). Maybe it's ok, because AuthToken comes directly from Django, but while testing it was kinda unusual(

PR #1144, opened 2023-02-23
Add callable default and options to app settings

### LINKED ISSUE #1050: Allow passing callable into project app settings options and default value
(**NOTE** from @mikkonie: exact spec for issue can be seen below in comments.)

### Problem

The project settings allow to define valid options for String value. This is nice if all possible values are known beforehand. However, if the options are to come from the database, this is not enough.

### Solution

Allow users to pass in a callable function that is passed the current `Project` object and returns a Django model choices like list. That is, a list-like of values being list-likes of length 2. The first entry is the value (e.g., a UUID), the second a label to display in the UI.

### Alternative Solutions

N/A

### Additional Context

- bihealth/varfish-server#782

gromdimon:
### Subtasks:
- [x] Added project and user to `AppSettingAPI`
- [x] Pass new kwargs where we use `get_default()` and `get_defaults()`
- [x] Add example app settings in `example_project_app` 
- [x] Test..test..test
- [x] Add kwargs `project=None`, `user=None` to `validate()`
- [x] Pass new kwargs in `set()`
- [x] Add check for callable options in `forms.py` under `_set_app_setting_widget()`
- [x] Implement support for lists of tuples returned from options in `_set_app_setting_widget()`
- [x] Pass to widget accordingly
- [x] Test tuple options

*Pretty complex*


example_project_app/plugins.py
```diff
@@ -19,6 +19,11 @@
 )
 
 
+def callable_function(project=None, user=None):
```
mikkonie: This could be renamed into something more descriptive, like `get_example_setting_default()`.

You should create one for each of the three settings scopes and make them utilize the arguments given. This would also make it easy to test these in unit tests.

Maybe make the type of these settings a string and return some meaningful string based on the arguments. Like, project title for a `PROJECT` scope setting, user name for `USER` scope setting, both for `PROJECT_USER` scope setting..

example_project_app/plugins.py
```diff
@@ -211,6 +216,24 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
             'default': '',
             'description': 'Example json project user setting',
         },
+        'project_callable_setting': {
+            'scope': SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT'],
+            'type': 'BOOLEAN',
+            'default': callable_function(),
+            'description': 'Example callable project setting',
```
mikkonie, line 272: For `PROJECT` and `USER` settings, I also recommend setting the `label` attribute. (for `PROJECT_USER` it is not needed, as those are not displayed in project/user forms)

example_project_app/plugins.py
```diff
@@ -211,6 +216,24 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
             'default': '',
             'description': 'Example json project user setting',
         },
+        'project_callable_setting': {
+            'scope': SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT'],
+            'type': 'BOOLEAN',
```
mikkonie: As mentioned in another comment, I recommend setting the type of these example settings to `STRING`.

projectroles/app_settings.py
```diff
@@ -35,7 +35,7 @@
     #:
     #:     'example_setting': {
     #:         'scope': 'PROJECT',  # PROJECT/USER
-    #:         'type': 'STRING',  # STRING/INTEGER/BOOLEAN
+    #:         'type': 'STRING',  # STRING/INTEGER/BOOLEAN/JSON
```
mikkonie, line 44: Good catch! :)

projectroles/app_settings.py
```diff
@@ -240,7 +240,9 @@ def _compare_value(cls, setting_obj, input_value):
         return setting_obj.value == str(input_value)
 
     @classmethod
-    def get_default(cls, app_name, setting_name, post_safe=False):
+    def get_default(
+        cls, app_name, setting_name, project=None, user=None, post_safe=False
```
mikkonie, line 268: Don't forget to add new args to the docstring.

projectroles/plugins.py
```diff
@@ -238,6 +238,16 @@ def revert_project_archive(
         # TODO: Implement this in your app plugin
         pass
 
+    def callable_function(self, project, user):
```
mikkonie: We can remove this from here, because this assumes a single function in every plugin. It'll be enough if we describe in the docs how these methods can be implemented.

example_project_app/plugins.py
```diff
@@ -19,6 +19,21 @@
 )
 
 
+def get_example_setting_default(project=None, user=None):
+    """Example callable function for different scopes
```
mikkonie: Just a style note here: Docstrings should have line break after the initial `"""`, also an empty line between the description and params/return value. Otherwise they will not render correctly when building API docs with Sphinx.

example_project_app/plugins.py
```diff
@@ -19,6 +19,21 @@
...
+    :param project: Project object
+    :param user: User object
+    :return: String with project and user info or 'No project'"""
+    response = 'Example callable setting'
```
mikkonie: This implementation looks good, but I think we should change the default response to something like "No project or user for callable". This makes it clear in the UI that the callable is working.

example_project_app/plugins.py
```diff
@@ -211,6 +226,60 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
...
+            'type': 'STRING',
+            'default': get_example_setting_default,
+            'description': 'Example callable project user setting',
+        },
+        'project_callable_setting_options': {
+            'scope': SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT'],
+            'type': 'STRING',
+            'label': 'Callable setting with options',
+            'default': get_example_setting_default,
+            'options': [
```
mikkonie: Here we should have just one method returning the list of options. Also, we should set up a separate example method to provide the example options. Note what I wrote in the spec, the return value must be a list of either direct values or tuples of `(value, label)`. Support for both must be implemented and tested, of course.

Please update this and related usage & tests.

projectroles/app_settings.py
```diff
@@ -260,6 +271,9 @@ def get_default(cls, app_name, setting_name, post_safe=False):
             app_settings = app_plugin.app_settings
 
         if setting_name in app_settings:
+            if callable(app_settings[setting_name].get('default')):
+                callable_setting = app_settings[setting_name].get('default')
+                return callable_setting(project, user)
```
mikkonie: One thing I forgot to mention in my previous review: whenever we call these callable methods, we should do it as follows:

- Call within `try/catch` block
- IF exception is returned, log it with `logger.error()`
    * AND set default value into something appropriate for the setting type (e.g. empty string for string, 0 for int, False for boolean, {} for json. You should define these in a dict in the `app_settings` module, if we haven't done so already)
- Same as above for options, but simply return one option which is the default value of the setting.

This avoids a situation where a bug in the implementation in a single callable from an app crashes the project/user form entirely and prevents its usage.

userprofile/forms.py
```diff
@@ -57,13 +57,24 @@ def __init__(self, *args, **kwargs):
...
-                            choices=[
-                                (option, option)
-                                for option in s_val.get('options')
-                            ],
-                            **setting_kwargs,
-                        )
+                        if callable(s_val.get('default')):
+                            self.fields[s_field] = forms.ChoiceField(
+                                choices=[
+                                    (str(option), str(option))
```
mikkonie: Something seems to be going wrong in passing the choices into the form :)
2023-03_form_option_bug

projectroles/tests/test_views.py
```diff
@@ -1809,6 +1828,8 @@ def test_post(self):
             'settings.%s.project_bool_setting' % EXAMPLE_APP_NAME: True,
             'settings.%s.project_json_setting'
             % EXAMPLE_APP_NAME: '{"Test": "Updated"}',
+            'settings.%s.project_callable_setting_options'
+            % EXAMPLE_APP_NAME: 'No project or user for callable',
```
mikkonie: While you're editing this part of the tests, you might want to remove these `% EXAMPLE_APP_NAME` format strings. I know you didn't add them and we've had them there for a while, but I just noticed this syntax and it seems to serve no purpose. It just looks confusing. I believe this is residue from some time where we thought of setting these tests up with multiple apps or something, but decided against it.

The example app is always the same, so we could write these out as simply `settings.example_project_app.setting_name`.

gromdimon:
There is a difficulty: After creation of Project appears another option to choose (look on screenshot), but if I choose it and try to save changes the following error appears:
And also the same thing happens with user settings(
```
Internal Server Error: /project/project/update/82aa54e8-4bff-469a-a693-d8893e05a8e7
Traceback (most recent call last):
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/projectroles/views.py", line 141, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/contrib/auth/mixins.py", line 104, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/edit.py", line 194, in post
    return super().post(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/edit.py", line 141, in post
    if form.is_valid():
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 175, in is_valid
    return self.is_bound and not self.errors
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 170, in errors
    self.full_clean()
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 373, in full_clean
    self._clean_form()
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 400, in _clean_form
    cleaned_data = self.clean()
  File "/home/gromdimon/Working/sodar-core/projectroles/forms.py", line 687, in clean
    if not self.app_settings.validate(
  File "/home/gromdimon/Working/sodar-core/projectroles/app_settings.py", line 626, in validate
    cls._check_value_in_options(
  File "/home/gromdimon/Working/sodar-core/projectroles/app_settings.py", line 183, in _check_value_in_options
    raise ValueError(
ValueError: Choice "('project', 'Hello world setting project')" not found in options (No project or user for callable)
Internal Server Error: /project/project/update/82aa54e8-4bff-469a-a693-d8893e05a8e7
Traceback (most recent call last):
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/projectroles/views.py", line 141, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/contrib/auth/mixins.py", line 104, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/edit.py", line 194, in post
    return super().post(request, *args, **kwargs)
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/views/generic/edit.py", line 141, in post
    if form.is_valid():
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 175, in is_valid
    return self.is_bound and not self.errors
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 170, in errors
    self.full_clean()
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 373, in full_clean
    self._clean_form()
  File "/home/gromdimon/Working/sodar-core/.venv/lib/python3.8/site-packages/django/forms/forms.py", line 400, in _clean_form
    cleaned_data = self.clean()
  File "/home/gromdimon/Working/sodar-core/projectroles/forms.py", line 687, in clean
    if not self.app_settings.validate(
  File "/home/gromdimon/Working/sodar-core/projectroles/app_settings.py", line 626, in validate
    cls._check_value_in_options(
  File "/home/gromdimon/Working/sodar-core/projectroles/app_settings.py", line 183, in _check_value_in_options
    raise ValueError(
ValueError: Choice "('project', 'Hello world setting project')" not found in options (No project or user for callable)
```
It looks like I have not updated options after creation of project or user model, so while validating the setting this new option just can not be found( But I am not sure


Screenshot:

Screenshot from 2023-03-06 17-03-37

<hr>

Also if the callable setting is already set to an emty string, than the default value isn't being shown while edditing it:

image

userprofile/forms.py
gromdimon: Actually It was correct, just my example function was too weird :)

example_project_app/plugins.py
```diff
@@ -19,6 +19,44 @@
...
+
+def get_example_setting_options(project=None, user=None):
+    """
+    Example callable function for options forming
+
+    :param project: Project object
+    :param user: User object
+    :return: List of strings
+    """
+    response = ['No project or user for callable']
```
mikkonie: I would like for the example method to return proper (value, label) tuples instead of a simple list, so it serves as a nice example for the potential users of the feature. You can e.g. use the UUID of the project/user as the actual value (with empty value for no project/user of course)

projectroles/app_settings.py
```diff
@@ -152,20 +158,40 @@ def _check_type_options(cls, setting_type, setting_options):
...
         :raise: ValueError if type is not recognized
         """
-        if setting_options and setting_value not in setting_options:
-            raise ValueError(
-                'Choice "{}" not found in options ({})'.format(
-                    setting_value, ', '.join(map(str, setting_options))
-                )
-            )
+        if setting_options:
+            if callable(setting_options):
```
mikkonie: Here we could limit redundant indentation by checking for setting_options and callable in the same if clause.

example_project_app/plugins.py
```diff
@@ -19,6 +19,45 @@
 )
 
 
+def get_example_setting_default(project=None, user=None):
+    """
+    Example callable function for different scopes
```
mikkonie: Very minor point, but note the missing periods in docstring descriptions :)

projectroles/tests/test_app_settings.py
```diff
@@ -253,6 +259,7 @@ class TestAppSettingAPI(
     """Tests for AppSettingAPI"""
 
     # NOTE: This assumes an example app is available
+    maxDiff = None
```
mikkonie: This we can remove from the final commit.

projectroles/forms.py
```diff
@@ -338,20 +338,48 @@ def _set_app_setting_widget(self, app_name, s_field, s_key, s_val):
...
-                        (int(option), int(option))
-                        if s_val['type'] == 'INTEGER'
-                        else (option, option)
-                        for option in s_val.get('options')
-                    ],
-                    **setting_kwargs
-                )
+                if callable(s_val.get('default')):
+                    if self.instance.pk:
+                        self.initial[s_field] = s_val.get('default')(
```
mikkonie: A small note, but since you already test the callable with `s_val.get('default')`, you don't have to use `get()` here anymore, the shorter syntax `s_val['default']` also works.

projectroles/forms.py
```diff
@@ -338,20 +338,48 @@ def _set_app_setting_widget(self, app_name, s_field, s_key, s_val):
...
+                                )
+                            ],
+                            **setting_kwargs
+                        )
+                    else:
+                        self.initial[s_field] = s_val.get('default')(
+                            project=None
+                        )
+                        self.fields[s_field] = forms.ChoiceField(
+                            choices=[
```
mikkonie, line 370: Something still seems to be going wrong here, we get an option where both the value and label are "Project UUID".

2023-03_form_option_bug2

projectroles/app_settings.py
```diff
@@ -260,6 +290,19 @@ def get_default(cls, app_name, setting_name, post_safe=False):
             app_settings = app_plugin.app_settings
 
         if setting_name in app_settings:
+            if callable(app_settings[setting_name].get('default')):
+                try:
+                    callable_setting = app_settings[setting_name].get('default')
```
mikkonie: Same as I've commented elsewhere, since we already check for `get('default')` earlier we can just use `app_setings[setting_name]['default']` here.

projectroles/forms.py
```diff
@@ -338,20 +338,48 @@ def _set_app_setting_widget(self, app_name, s_field, s_key, s_val):
...
-                )
+                if callable(s_val.get('default')):
+                    if self.instance.pk:
+                        self.initial[s_field] = s_val.get('default')(
+                            project=self.instance
+                        )
+                        self.fields[s_field] = forms.ChoiceField(
+                            choices=[
+                                (str(option), str(option))
+                                for option in s_val.get('options')(
```
mikkonie: I was unable to find the check for whether `options` is a callable. There only seems to be that check for `default`? Default and options do not depend on each other and either, neither or both can be set as callable.

Possibly we didn't catch this because your example settings with a callable `option` also have a callable `default`, so this condition is always reached?

userprofile/forms.py
```diff
@@ -57,13 +57,24 @@ def __init__(self, *args, **kwargs):
...
+                                    for option in s_val.get('options')(
```
mikkonie: Same comment here as in the project form, I can't find a check for whether `options` are callable.

projectroles/forms.py
gromdimon: To be honest, I still do not understand what is designed to be here( Befire there were implemented tuples (as was described in previous review), but now I changed example callable option method to a proper string. Could you somehow specify what exact tuples should be there and how it should look like?

example_project_app/plugins.py
mikkonie: These are strings again, please return proper value/label tuples.

projectroles/forms.py
mikkonie: The way HTML options work is that there is a value and a label. Django obviously supports that in its forms as well. `<option value="value">label</option>`. As I have previously specified, we should support both:

- If the method returns a list of strings, use that
- If the method returns a list of tuples for value and label, provide that to Django.

The latter is of course the nicer method, so the example implementation should also provide nice tuples for the form with the exact value as value and a human readable label as the label.

projectroles/app_settings.py
```diff
@@ -260,7 +285,24 @@ def get_default(cls, app_name, setting_name, post_safe=False):
             app_settings = app_plugin.app_settings
 
         if setting_name in app_settings:
-            if app_settings[setting_name]['type'] == 'JSON':
+            if callable(app_settings[setting_name].get('default')):
+                try:
+                    callable_setting = app_settings[setting_name]['default']
+                    if app_settings[setting_name].get('options'):
```
mikkonie: There seems to be a mistake here. The callable default value and callable options should be handled separaterly. Here we seem to get the default value using a callable if `options` are set in the app settings definition. These should not be related. Or am I misundestanding something?

Here's how it's supposed to work:

- If `default` is a callable, call the default callable to get the default value
- If `options` is a callable, call the options callable to get the options.

There should be no direct relationship between the two. Both, either or neither of them could be a callable for a specific app setting.

userprofile/forms.py
gromdimon: I think here is problem in Github UI)) I've changed it to tuples, see the code section

example_project_app/plugins.py
gromdimon: I think here is problem in Github UI)) I've changed it to tuples, see the code section

projectroles/app_settings.py
gromdimon: The thing here is, that in exmple setting I have following:
```
'project_callable_setting_options': {
            'scope': SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT'],
            'type': 'STRING',
            'label': 'Callable setting with options',
            'default': get_example_setting_options,
            'options': get_example_setting_options,
            'description': 'Example callable project setting with options',
            'user_modifiable': True,
        }
```
In that case, when I am passing default value for setting with options, I need to pass option functios, that will return a default value. There was before `get_example_setting_default` function, but as you said it has to be independant with option function.

example_project_app/plugins.py
```diff
@@ -19,6 +19,59 @@
...
+    """
+    Example callable function for options forming.
+
+    :param project: Project object
+    :param user: User object
+    :return: List of tuples for ChoiceField
+    """
+    response = [
+        ('N/A', 'No project or user for callable'),
+        'Example string option',
```
gromdimon, line 50: Here is check for string

projectroles/forms.py
```diff
@@ -338,20 +338,47 @@ def _set_app_setting_widget(self, app_name, s_field, s_key, s_val):
...
-                        for option in s_val.get('options')
-                    ],
-                    **setting_kwargs
-                )
+                if callable(s_val['options']):
+                    if self.instance.pk:
+                        values = s_val['options'](project=self.instance)
+                        print(values)
+                        self.fields[s_field] = forms.ChoiceField(
+                            choices=[
```
gromdimon, line 359: Here's field

PR #1153, opened 2023-03-01
Projectroles/features

### LINKED ISSUE #1141: Remove scrolling from project list in category details view
This seemed like a good idea at the time, but in the end is just confusing, not very usable in large categories and doesn't fit the overall conventions. Not just my own feeling, I have received similar feedback from users.

Filtering should still remain though. We should just display the entire table instead of making it scroll.

## Spec

- Remove `sodar-pr-project-list-body-parent` use from `_project_list.html`
- Remove `sodar-pr-project-list-body-parent` definition from `projectroles.css`
- (No testing necessary, just ensure the list looks as expected under a category :))

### LINKED ISSUE #1151: Investigate/fix project form JSON app setting initial value passing
I think I missed this in an earlier review, but [in recent changes to projectroles forms](https://github.com/bihealth/sodar-core/blob/d72a39d0b0699a91891211f39149cd4c945be9e1/projectroles/forms.py#L634), the following can be found:

```
# for some reason, there is a distinct possibility, that the
# initial value has been discarded and we get '' as value.
# Seems to only happen in automated tests. Will catch that
# here.
if not self.cleaned_data.get(s_field):
    self.cleaned_data[s_field] = '{}'
```

Generally, if something only happens in unit tests and can't be replicated in normal use, something is wrong with the tests.

Look into which tests specifically fail and how they are set up. If it's a POST request, maybe the initial value is not correctly provided in the request?

It is possible my hunch is wrong and this can also occur in actual use. But even so, we should know how and why.

gromdimon:
Issue #1151 #1141

### DIFF
--- a/projectroles/forms.py
+++ b/projectroles/forms.py
@@ -631,21 +631,11 @@ class ProjectForm(SODARModelForm):
                     s_field = 'settings.{}.{}'.format(name, s_key)
 
                     if s_val['type'] == 'JSON':
-                        # for some reason, there is a distinct possibility, that the
-                        # initial value has been discarded and we get '' as value.
-                        # Seems to only happen in automated tests. Will catch that
-                        # here.
                         if not self.cleaned_data.get(s_field):
                             self.cleaned_data[s_field] = '{}'
-                        try:
-                            self.cleaned_data[s_field] = json.loads(
-                                self.cleaned_data.get(s_field)
-                            )
-                        except json.JSONDecodeError as err:
-                            # TODO: Shouldn't we use add_error() instead?
-                            raise forms.ValidationError(
-                                'Couldn\'t encode JSON\n' + str(err)
-                            )
+                        self.cleaned_data[s_field] = json.loads(
+                            self.cleaned_data.get(s_field)
+                        )
                     elif s_val['type'] == 'INTEGER':
                         # When the field is a select/dropdown the information of
                         # the datatype gets lost. We need to convert that here,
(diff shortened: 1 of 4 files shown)


projectroles/forms.py
```diff
@@ -631,21 +631,11 @@ def clean(self):
...
-                        # initial value has been discarded and we get '' as value.
-                        # Seems to only happen in automated tests. Will catch that
-                        # here.
                         if not self.cleaned_data.get(s_field):
                             self.cleaned_data[s_field] = '{}'
-                        try:
-                            self.cleaned_data[s_field] = json.loads(
-                                self.cleaned_data.get(s_field)
-                            )
-                        except json.JSONDecodeError as err:
```
mikkonie, line 644: We still need to catch possible JSON encoding errors, because it is completely possible for the user to input invalid JSON data.

gromdimon: I've investigated this bug and found out, that we need check for empty string in `self.cleaned_data[s_field]` in case of empty value in form. Check for valid JSON string is also needed as you mentioned above.
The only thing is that I changed `forms.ValidationError` to `self.add_error`. As a result there is no more a banner on the top of form, but a small hint near JSON field. Idk if I did it right, let me know.

gromdimon:
Also noticed, that in UserForm there is similar bug, so I made corresponding chnages.

projectroles/forms.py
mikkonie: Indeed, `add_error()` is what we typically should use. I have no idea why the exception was raised here, possibly just some mistake on my part.

It also seems our override of `add_error()` does something parsing if it receives `ValidationError` as argument instead of string. I believe it's a formatting thing when an exception contains multiple messages. In this case that probably isn't relevant as we format the error message before the call.

So yes, this looks valid to me :)

PR #1154, opened 2023-03-02
Docs update

### LINKED ISSUE #860: Add plugin hook for validating project update form for app settings
There are a few cases in SODAR where we would need to do some more complex validation for user-provided app settings, for example bihealth/sodar-server#1310. Also, in some cases we might have other special rules for e.g. syntax and project context, which can't be represented by the setting definition dict alone.

To tackle such issues, we should add a method to project app plugins for validating app settings in the project update form. The method would receive all the settings of that particular app and return errors for individual settings if found.

## Spec

- Add skeleton and docstring of `validate_form_app_settings()` method to `ProjectAppPluginPoint`
    * Args:
        + `project` (`Project` object or `None`)
        + `user` (`User` object or `None`)
        + `app_settings` (dict in format of {'setting_name': 'value')
     * Return value = dict in format of `{'setting_name': 'Error string' or None}`
         + The skeleton should simply return `None` to signify "not implemented"
- Add similar method skeleton to `SiteAppPluginPoint`
    * Same signature *without* the `project` arg
- In `ProjectForm` and `UserForm`, call validation
    * Within `clean()`
    * Call method in each plugin per app, only provide the app's own settings to the method
    * If the method returns `None`, the app's settings are considered to be OK
    * IF setting name is found in return dict and the error value is not `None`, THEN add error to form errors for the related field
- Add example implementation of method to `example_project_app`
    * For both `PROJECT` and `USER` scopes (`PROJECT_USER` is exempt here as those are not included in forms)
- Add tests where these pass/fail
- Update docs where applicable

## Tasks

- [x] Add method skeletons to be implemented
- [x] Add usage in forms
- [x] Add example implementation
- [x] Add tests where example implementation passes/fails
- [ ] Update docs

gromdimon:
### Tasks:

- [x] Add method skeletons to be implemented
- [x] Add usage in forms
- [x] Add example implementation
- [x] Add tests where example implementation passes/fails
- [x] Update docsa

<hr>
p.s. I've also updated some docstrings in `ProjectForm` as I found it a bit confusing the first time, probably it is a good idea. Let me know if it doesn't look good!

### DIFF
--- a/example_project_app/plugins.py
+++ b/example_project_app/plugins.py
@@ -275,3 +275,26 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
                     action=action.lower(),
                 ),
             )
+
+    def validate_form_app_settings(self, app_settings, project=None, user=None):
+        """Example implementation for app setting validation"""
+        errors = {}
+        if (
+            app_settings.get('scope')
+            == SODAR_CONSTANTS['APP_SETTING_SCOPE_USER']
+            and not user
+        ):
+            errors[
+                app_settings['name']
+            ] = 'No user provided for user scope setting'
+        if (
+            app_settings.get('scope')
+            == SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT']
+            and not project
+        ):
+            errors[
+                app_settings['name']
+            ] = 'No project provided for project scope setting'
+        if errors == {}:
+            return None
+        return errors
--- a/example_project_app/plugins.py
+++ b/example_project_app/plugins.py
@@ -279,22 +279,23 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
     def validate_form_app_settings(self, app_settings, project=None, user=None):
         """Example implementation for app setting validation"""
         errors = {}
-        if (
-            app_settings.get('scope')
-            == SODAR_CONSTANTS['APP_SETTING_SCOPE_USER']
-            and not user
-        ):
-            errors[
-                app_settings['name']
-            ] = 'No user provided for user scope setting'
-        if (
-            app_settings.get('scope')
-            == SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT']
-            and not project
-        ):
-            errors[
-                app_settings['name']
-            ] = 'No project provided for project scope setting'
+        for setting in app_settings:
+            if (
+                app_settings[setting].get('scope')
+                == SODAR_CONSTANTS['APP_SETTING_SCOPE_USER']
+                and not user
+            ):
+                errors[
+                    app_settings['name']
+                ] = 'No user provided for user scope setting'
+            elif (
+                app_settings[setting].get('scope')
+                == SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT']
+                and not project
+            ):
+                errors[
+                    app_settings['name']
+                ] = 'No project provided for project scope setting'
         if errors == {}:
             return None
         return errors
--- a/example_project_app/plugins.py
+++ b/example_project_app/plugins.py
@@ -275,26 +275,3 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
                     action=action.lower(),
                 ),
             )
-
-    def validate_form_app_settings(
-        self, app_settings, setting_name, project=None, user=None
-    ):
-        """Example implementation for app setting validation"""
-        errors = {}
-        if (
-            app_settings[setting_name].get('scope')
-            == SODAR_CONSTANTS['APP_SETTING_SCOPE_USER']
-            and not user
-        ):
-            errors[setting_name] = 'No user provided for user scope setting'
-        elif (
-            app_settings[setting_name].get('scope')
-            == SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT']
-            and not project
-        ):
-            errors[
-                setting_name
-            ] = 'No project provided for project scope setting'
-        if errors == {}:
-            return None
-        return errors
--- a/example_project_app/plugins.py
+++ b/example_project_app/plugins.py
@@ -276,26 +276,25 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
                 ),
             )
 
-    def validate_form_app_settings(self, app_settings, project=None, user=None):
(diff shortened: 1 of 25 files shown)


gromdimon:
I found out, that there were some typos in docs, so I improved that as well :)

example_project_app/plugins.py
```diff
@@ -275,3 +275,26 @@ def perform_project_modify(
                     action=action.lower(),
                 ),
             )
+
+    def validate_form_app_settings(self, app_settings, project=None, user=None):
+        """Example implementation for app setting validation"""
+        errors = {}
+        if (
+            app_settings.get('scope')
```
mikkonie: This example is nice, but there seems to be a mistake here. The method is supposed to receive all the app settings related to the app (plugin). each of those settings have their own scope, name etc, so this should be something like `app_settings[setting_name]['scope']` instead.

projectroles/forms.py
```diff
@@ -312,7 +312,14 @@ def _get_parent_choices(cls, instance, user):
...
     def _set_app_setting_widget(self, app_name, s_field, s_key, s_val):
-        """Internal helper for setting app setting widget and value"""
+        """
+        Internal helper for setting app setting widget and value
+
+        :param app_name: App name
+        :param s_field: Form field name
+        :param s_key: Setting key
+        :param s_val: Setting value
+        """
```
mikkonie: I appreciate writing out these docstrings in total. For internal helpers it's not necessarily critical, but does help understanding what's happening in development. I do try to write these out myself nowadays. Only one very tiny note, in case of a "full" docstring like this it is convention to add a period `.` after the description :)

projectroles/forms.py
```diff
@@ -541,9 +554,80 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
             self.fields['description'].widget = forms.HiddenInput()
             self.fields['readme'].widget = forms.HiddenInput()
 
+    @classmethod
+    def _app_settings_validation(
```
mikkonie: It is usually convention to name functions/methods as a verb, so something like `_validate_app_settings()` would be more appropriate.

projectroles/forms.py
```diff
@@ -541,9 +554,80 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
...
+        cleaned_data,
+        app_plugins,
+        app_settings,
+        p_kwargs,
+        instance,
+        instance_owner_as,
+    ):
+        """Validate and clean app_settings form fields"""
+        errors = []
+        if cleaned_data.get('type') == PROJECT_TYPE_PROJECT:
```
mikkonie: To avoid extreme indentation, we could also just return here with `cleaned_data` if type is `PROJECT_TYPE_CATEGORY`.

projectroles/plugins.py
```diff
@@ -608,6 +620,17 @@ def get_extra_data_link(self, _extra_data, _name):
         # TODO: Implement this in your app plugin
         return None
 
+    def validate_form_app_settings(self, app_settings, user=None):
```
mikkonie: Only project and site apps are expected to provide app settings, so for `BackendPluginPoint` this is not needed.

gromdimon: But this `validate_form_app_settings` is for `SiteAppPluginPoint`. The `BackendPluginPoint` has no such method

projectroles/forms.py
```diff
@@ -541,9 +554,83 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
...
+        if cleaned_data.get('type') == PROJECT_TYPE_CATEGORY:
+            return (cleaned_data, errors)
+
+        for plugin in app_plugins + [None]:
+            if plugin:
+                name = plugin.name
+                p_settings = app_settings.get_definitions(
+                    APP_SETTING_SCOPE_PROJECT, plugin=plugin, **p_kwargs
+                )
+                app_settings_errors = plugin.validate_form_app_settings(
```
mikkonie: Here, you provide the definitions of app settins to the validation method. However, what we actually need to provide is the list of actual settings and their values, from `cleaned_data`. Also, only the settings of the specific app plugin should be provided to the appropriate plugin.

Of course, the plugin already knows about its own definitiions, so those do not have to be provided to this method.

So, to clarify:

- For each app plugin, get the settings values for `settings.{APP_NAME}.{SETTING_NAME}`
    * For settings defined in the projectroles app, we can skip this
- Provide that as the `app_settings` dict to the appropriate app's method, with setting name as key
    * Strip the setting name to only contain the last part, the other info is only interesting to the form
- The function is expected to find related values by key which is the setting name

projectroles/forms.py
```diff
@@ -541,9 +554,88 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
...
+                                errors.append((field, error))
+                except AttributeError:
+                    # Plugin does not have a validate_app_settings method
+                    pass
+
+                if s_val['type'] == 'JSON':
+                    # for some reason, there is a distinct possibility, that the
+                    # initial value has been discarded and we get '' as value.
+                    # Seems to only happen in automated tests. Will catch that
+                    # here.
```
mikkonie: I think you just fixed this in another PR by not raising `ValidationError`? Here we still seem to have the old code in.

example_project_app/plugins.py
```diff
@@ -275,3 +275,26 @@ def perform_project_modify(
                     action=action.lower(),
                 ),
             )
+
+    def validate_form_app_settings(
+        self, app_settings, setting_name, project=None, user=None
```
mikkonie: There should not be a `setting_name` argument here. As specified before, this method should receive all of the app's app settings as one dict and validate one, several or all settings within that one method call. This allows us to e.g. validate multiple settings against each other, which is one of the use cases for this feature.

projectroles/forms.py
```diff
@@ -541,9 +554,88 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
...
+                p_settings = app_settings.get_definitions(
+                    APP_SETTING_SCOPE_PROJECT, app_name=name, **p_kwargs
+                )
+
+            for s_key, s_val in p_settings.items():
+                s_field = 'settings.{}.{}'.format(name, s_key)
+
+                try:
+                    app_settings_errors = plugin.validate_app_settings(
+                        p_settings,
```
mikkonie: We are still providing app setting definitions here, when we should instead be providing the actual values input by the user into the form so they can be validated. The definitions are known by the plugin so they do not need to be provided to this method. Please go back and read what I wrote in the previous review.

PR #1158, opened 2023-03-07
Disable app setting updates for categories

### LINKED ISSUE #1155: Category update timeline event includes app settings
When updating a category, the `project_update` timeline event lists app settings as changed values, even though we hide them in the form and do not allow updating them at all.

Project update timeline event seems to work fine. I suspect this is related to the changes made in #20.

To reproduce, simply edit something (e.g. description or readme) in a category and observe the timeline event list.

App settings should not be included in the timeline event description or extra data if updating a category. 

2023-03_category_timeline_bug

### LINKED ISSUE #1159: Add migration to clean up category app settings
With #20 and #1151, we are explicitly not allowing app settings for categories. However, prior to implementing aforementioned tasks, project app settings for categories were still created and updated upon category modification. They just were never used.

We should add a migration to `projectroles` to clean up app settings if `project` exists and `project__type==PROJECT_TYPE_CATEGORY`.

I initially thought this could be an addition to the `cleanappsettings` command, but it's better to delete these outright as a migration.

### LINKED ISSUE #1162: Prevent creating app settings for categories
Since we are currently forbidding creating app settings for categories, this should also be prevented in the model and API.

## Spec

- Add validation method to `AppSetting` model
    * If `project` exists and `project__type=PROJECT_TYPE_CATEGORY`, raise exception
- Add similar check to `AppSettingAPI.set()`
- Add tests
- Update docs to reflect that app settings can not be linked to categories

gromdimon:
### ToDo:
- [x] Reform Forms and Views
- [x] Update tests if needed


mikkonie:
This seems to work perfectly now, good work! Please make sure view tests for project creation and update ensure settings are created/updates for projects but not categories.

gromdimon:
## Another Issue #1162 
### Tasks:
- [x] Add validation method to AppSetting model
- [x] Add similar check to AppSettingAPI.set()
- [x] Add tests
- [x] Update docs to reflect that app settings can not be linked to categories

gromdimon:
### Also Migration from #1159 
- [x] ?

projectroles/migrations/0026_delete_category_settings.py
```diff
@@ -0,0 +1,26 @@
...
+def clean_up_app_settings(apps, schema_editor):
+    Project = apps.get_model('projectroles', 'Project')
+    AppSetting = apps.get_model('projectroles', 'AppSetting')
+
+    # Get all projects that are of type PROJECT_TYPE_CATEGORY
+    category_projects = Project.objects.filter(type='CATEGORY')
+
+    # Loop through all category projects and delete their app settings
+    for project in category_projects:
+        app_settings = AppSetting.objects.filter(project=project)
```
mikkonie: We can simply do `filter(project__type=PROJECT_TYPE_CATEGORY)` here, so we can avoid having to query for Project objects separately.

projectroles/app_settings.py
```diff
@@ -426,6 +427,15 @@ def _log_debug(action, app_name, setting_name, value, project, user):
...
                 q_kwargs['app_plugin__name'] = app_name
             setting = AppSetting.objects.get(**q_kwargs)
+
+            # Check if setting is set to a category
+            if (
+                setting
+                and setting.project
+                and setting.project.type == PROJECT_TYPE_CATEGORY
+            ):
+                return False
```
mikkonie: Instead of returning False, I suggest we instead raise a `ValueError` with a message about settings not being allowed for categories. That makes it harder to ignore if a user attempts to misuse the feature :)

This check and potential exception raising should be in the beginning of `set()` along with the check for project and user args. This also means we only need to check this once and can remove the latter check under ` except AppSetting.DoesNotExist:`.

gromdimon:
## Finally Issue #1169

PR #1161, opened 2023-03-10
Add app settings custom validation

gromdimon:
Previous PR: #1154

### DIFF
--- a/example_project_app/plugins.py
+++ b/example_project_app/plugins.py
@@ -279,21 +279,14 @@ class ProjectAppPlugin(ProjectModifyPluginMixin, ProjectAppPluginPoint):
     def validate_form_app_settings(self, app_settings, project=None, user=None):
         """Example implementation for app setting validation plugin API"""
         errors = {}
-        for setting_name in app_settings:
+        for setting_name, setting_value in app_settings.items():
             if (
-                app_settings[setting_name].get('scope')
-                == SODAR_CONSTANTS['APP_SETTING_SCOPE_USER']
-                and not user
-            ):
-                errors[setting_name] = 'No user provided for user scope setting'
-            elif (
-                app_settings[setting_name].get('scope')
-                == SODAR_CONSTANTS['APP_SETTING_SCOPE_PROJECT']
-                and not project
+                setting_name == 'project_hidden_setting'
+                and setting_value == 'Example project hidden setting'
             ):
                 errors[
                     setting_name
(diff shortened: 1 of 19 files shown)


projectroles/forms.py
```diff
@@ -541,9 +553,86 @@ def __init__(self, project=None, current_user=None, *args, **kwargs):
...
+                ):
+                    errors.append((s_field, 'Invalid value'))
+
+            # Custom validation for app settings
+            try:
+                app_settings_errors = plugin.validate_app_settings(
+                    plugin_app_settings,
+                    project=instance,
+                    user=instance_owner_as,
+                )
```
gromdimon, line 620: Here I pass in `plugin.validate_app_settings` such data:
for `bgjobs` - {}
for `example_project_app`  - {'project_str_setting': '', 'project_int_setting': 0, 'project_str_setting_options': 'string1', 'project_int_setting_options': '0', 'project_bool_setting': True, 'project_global_setting': False, 'project_json_setting': '{"list": [1, 2, 3, 4, 5], "Example": "Value", "level_6": false}', 'project_hidden_setting': '', 'project_hidden_json_setting': '{}'}
for `filesfolders` - {'allow_public_links': False}
for `timeline` - {}
and for `projectroles` - {'ip_restrict': False, 'ip_allowlist': '{}'}

Is it what supposed to be?

mikkonie: That looks exactly correct :)

example_project_app/plugins.py
```diff
@@ -275,3 +275,25 @@ def perform_project_modify(
                     action=action.lower(),
                 ),
             )
+
+    def validate_form_app_settings(self, app_settings, project=None, user=None):
+        """Example implementation for app setting validation plugin API"""
+        errors = {}
+        for setting_name in app_settings:
```
mikkonie: Now, since you are correctly providing the values provided by the user in the form here, you should also do some checks on the values in this example implementation. Then test against some unaccepted values in your unit tests.

example_project_app/plugins.py
```diff
@@ -275,3 +275,25 @@ def perform_project_modify(
...
+            if (
+                app_settings[setting_name].get('scope')
```
mikkonie: Of course, definitions such as scope will not be available in this value now because you are simply providing the value. But that's ok, we don't necessarily have to validate against those here. Just come up with some rules for the example project app values and check those in test :)

PR #1165, opened 2023-03-13
Projectroles/sidebar tags to context processor

### LINKED ISSUE #969: Replace sidebar template tags with a context processor
Most of these template tags only get called from the base template and this happens every time we load a UI view. Also, they do not depend on any data on a specific view/template themselves. Hence, having them as template tags is unnecessary. Also, having this as a single context processor will most likely perform better. And albeit unlikely, this would give a site better control over customizing the sidebar.

This is a breaking change as it will require SODAR Core based sites to add a context processor. Hence this must be mentioned in the major changes doc.

## Spec

- Create new `sidebar_processor` under `projectroles.context_processors`
    * Don't forget to add this under `context_processors` in `base.py` config :)
- Move the code of the following template tags from `projectroles_tags` into the context processor:
    * `get_sidebar_icon_size()`
    * `get_sidebar_notch_pos()`
    * `get_sidebar_notch_size()`
    * `get_sidebar_padding()`
- **NOTE:** `get_sidebar_app_legend()` should remain as a template tag, as the output is dynamic based on app/template content
- Update tests
    * Remove related tests from `projectroles.tests.test_templatetags`
    * Instead, test for the expected values in view context in a test in `projectroles.tests.test_views`.
- Update docs
    * Add new context processor in `app_projectroles_settings.rst`
    * Add note about this new processor being required under "breaking changes" for the latest release in `major_changes.rst`

## Takss

- [ ] Create and enable new context processor
- [ ] Move functionality from template tags into context processor
- [ ] Update tests
- [ ] Update docs

gromdimon:
### Tasks:
- [x] Create and enable new context processor
- [x] Move functionality from template tags into context processor
- [x] Update tests
- [x] Update docs


docs/source/major_changes.rst
```diff
@@ -49,6 +49,17 @@ Release Highlights
...
+New Context Processors
+------------------
+
+Several templatetags have been moved to context processors. If you have
+customized the following tags, you need to update it's usage:
+
+- ``projectroles_tags.get_sidebar_icon_size()``
+- ``projectroles_tags.get_sidebar_notch_pos()``
+- ``projectroles_tags.get_sidebar_notch_size()``
+- ``projectroles_tags.get_sidebar_padding()``
```
mikkonie: These are only ever used by projectroles. The important thing we need to tell the user here is to include `projectroles.context_processors.sidebar_processor'` in `TEMPLATES['OPTIONS']['context_processors']` in the base config of their Django site.

Also note the hanging underlining of the title. Sphinx is very strict about these, sadly :)

projectroles/context_processors.py
```diff
@@ -48,3 +56,38 @@ def app_alerts_processor(request):
...
+        """Return sidebar icon size with a min/max limit"""
+        return sorted(
+            [
+                SIDEBAR_ICON_MIN_SIZE,
+                getattr(settings, 'PROJECTROLES_SIDEBAR_ICON_SIZE', 32),
+                SIDEBAR_ICON_MAX_SIZE,
+            ]
+        )[1]
+
+    def get_sidebar_notch_pos():
```
mikkonie: We don't need to have separate functions for these, we can simply return them in the return value dict instead. Maybe the `sidebar_icon_size` we should populate in a separate variable first to avoid ugly nesting in code.

projectroles/tests/test_views.py
```diff
@@ -62,6 +62,7 @@
     MSG_INVITE_USER_NOT_EQUAL,
     MSG_INVITE_USER_EXISTS,
 )
+from projectroles.templatetags import projectroles_tags as tags
```
mikkonie: This import should not be needed here, at least not for this feature.

projectroles/tests/test_views.py
```diff
@@ -182,6 +183,81 @@ def test_render_anon(self):
...
         self.assertEqual(response.status_code, 200)
 
+    def test_context_get_sidebar_icon_size(self):
+        """Test getting sidebar icon size in context"""
+        with self.login(self.user):
+            response = self.client.get(reverse('home'))
+            self.assertEqual(response.context['sidebar_icon_size'], 36)
+
+    @override_settings(
+        PROJECTROLES_SIDEBAR_ICON_SIZE=tags.SIDEBAR_ICON_MIN_SIZE - 2
```
mikkonie: We should include these constants from `context_processors` and remove them from `tags`.

PR #1174, opened 2023-03-15
Project type scope setting

### LINKED ISSUE #1169: Add optional project type scope to app settings
While doing #1162 as I instructed, @gromdimon noticed, quite rightly, that after recent changes project/category starring is an app setting and it must also work for categories.

Disabling app settings for categories was originally going to be a temporary step before we wrote proper allowing/disallowing of project types for app settings definitions. So let's just go directly to that more advanced step.

## Spec

- Add optional `project_types` field to app settings definitions
    * Expected value is a list which can contain `PROJECT_TYPE_PROJECT`, `PROJECT_TYPE_CATEGORY` or both
    * If value is not set, assume it to be `[PROJECT_TYPE_PROJECT]`
- Update `AppSettingAPI.set()`
    * IF `project` arg is set
        + AND `project_types` for setting is unset and `project.type != PROJECT_TYPE_CATEGORY`
        + OR `project_types` is set and `project.type not in project_types`
    * THEN raise `ValueError`
- Set the project starring setting `project_types` to accept both types, of course :)
- In project create/update form, display/set only app settings where the project type matches the setting definition
    * That is, for categories, nothing of the currently existing app settings should be displayed or set
    * You can add an example "project category setting" to be displayed and tested against
- Add migration (or update the current one in your work branch) to clean up app settings
    * FOR each existing app setting
        + IF `project` is set and its type doesn't match `project_types` for that app setting, delete object
- Add tests, of course :)

gromdimon:
### Tasks:
- [x] Implement
- [x] Test
- [x] Add migration
- [x] Documentation


projectroles/forms.py
```diff
@@ -322,6 +322,19 @@ def _set_app_setting_widget(self, app_name, s_field, s_key, s_val):
...
+        s_project_types = s_val.get('project_types') or [PROJECT_TYPE_PROJECT]
+        if PROJECT_TYPE_PROJECT in s_project_types:
+            s_widget_attrs['data-project-types'] = 'project'
+        elif PROJECT_TYPE_CATEGORY in s_project_types:
+            s_widget_attrs['data-project-types'] = 'category'
+        elif (
+            PROJECT_TYPE_PROJECT in s_project_types
+            and PROJECT_TYPE_CATEGORY in s_project_types
+        ):
+            s_widget_attrs['data-project-types'] = 'project,category'
```
mikkonie: I think we can simplify this by setting the attr like this: `s_widget_attrs['data-project-types'] = ','.join(s_project_types)`

Of course this will result in the strings "PROJECT" and "CATEGORY", so in the JQuery we have to check against the uppercase strings.

projectroles/migrations/0026_delete_category_settings.py
```diff
@@ -0,0 +1,47 @@
...
+from django.db import migrations
+from projectroles.app_settings import AppSettingAPI
+
+
+def clean_up_app_settings(apps, schema_editor):
+    app_settings_api = AppSettingAPI()
+    AppSetting = apps.get_model('projectroles', 'AppSetting')
+
+    # Find all app settings with project type 'CATEGORY'
+    app_settings = AppSetting.objects.filter(project__type='CATEGORY')
```
mikkonie: I believe what we want to do here is:

- Query for all `AppSetting` objects where `project != None`
- Compare the project type against the app setting definition for that setting
    * IF `project.type not in setting_def.project_types`
        + OR `not setting_def.get('project_types') AND project.type == PROJECT_TYPE_CATEGORY`
        + OR definition is not found at all
    * THEN delete `AppSetting` object

projectroles/migrations/0026_delete_category_settings.py
```diff
@@ -0,0 +1,47 @@
...
+
+    for app_setting in app_settings:
+        try:
+            setting = app_settings_api.get_definition(
+                app_name=app_setting.app_plugin.name, name=app_setting.name
+            )
+        except ValueError:
+            # AppSetting is already deleted
```
mikkonie: This is not exactly correct. `get_definition()` only returns the dict from a plugin (or projectroles). It is possible for "orphaned" `AppSetting` objects to remain in the database if a definition has been removed from the plugin/projectroles dict. The `cleanappsettings` management command usually takes care of these.

mikkonie:
One more thing: please ensure that the REST API also checks for the correct project type. 

- Add a test in `TestProjectSettingSetAPIView` to ensure attempting to set a setting value for the wrong project type returns `HTTP 400`
- This **should** already work due to existing validation in the app settings API..
- ..but if not, add the required logic to `ProjectSettingSetAPIView`

projectroles/forms.py
gromdimon: Just hilarious how i overcomplicated this :)

projectroles/migrations/0026_delete_category_settings.py
gromdimon: But in that case I receive such an error:
ValueError: App setting not found in app "example_project_app" with name "project_category_str_setting"
I think as ValueError can be raised in get_definition(), we have to have this exception.

projectroles/migrations/0026_delete_category_settings.py
```diff
@@ -0,0 +1,46 @@
...
+
+from django.db import migrations
+from projectroles.app_settings import AppSettingAPI
+
+
+def clean_up_app_settings(apps, schema_editor):
+    app_settings_api = AppSettingAPI()
+    AppSetting = apps.get_model('projectroles', 'AppSetting')
+
+    # Find all app settings with project type 'CATEGORY'
```
mikkonie: This comment needs to be updated ;)

projectroles/migrations/0026_delete_category_settings.py
```diff
@@ -0,0 +1,46 @@
...
+    app_settings = AppSetting.objects.exclude(project=None)
+
+    for app_setting in app_settings:
+        try:
+            setting_def = app_settings_api.get_definition(
+                app_name=app_setting.app_plugin.name, name=app_setting.name
+            )
+        except ValueError:
+            continue
```
mikkonie, line 26: We should also delete the AppSetting if no definition is found. Also, we can simplify this by doing just one `get_definition()` call and first set `app_name` as follows:

`app_name = app_setting.app_plugin.name if app_setting.app_plugin else 'projectroles'`

projectroles/templates/projectroles/project_form.html
```diff
@@ -57,7 +57,7 @@ <h2>Create {{ project_display }} or {{ category_display }} Under {{ parent.title
     {% elif disable_categories %}
       <h2>Create {{ project_display }}</h2>
     {% else %}
-      <h2>Create Top Level {{ category_display }}</h2>
+      <h2 id="object-type">Create Top Level {{ category_display }}</h2>
```
mikkonie: Comparing against freely entered strings is always a bit risky. I'd recommend instead to set e.g. a `data-project-action` attribute here and set some short descriptor there. That way, we avoid the risk of the JQuery breaking if we decide to reword the displayed title for the view.

projectroles/migrations/0026_delete_category_settings.py
mikkonie: My comment above is not yet fully handled: As I wrote: "We should also delete the AppSetting if no definition is found." Right now, if we don't find the setting definition, we simply `continue` to the next iteration.

projectroles/templates/projectroles/project_form.html
```diff
@@ -38,7 +38,7 @@
   <div class="row sodar-subtitle-container">
     {% if object.pk %}
       {# NOTE: h3 is correct here, as we also have the project header #}
-      <h3 id="object-type">Update {% get_display_name object.type title=True %}</h3>
+      <h3 id="object-type" data-project-action={{ object.type }}>Update {% get_display_name object.type title=True %}</h3>
```
mikkonie: If you still want to have the `id` attribute here, I suggest naming it along with the usual naming conventions. `sodar-pr-project-form-title` would work.

PR #1181, opened 2023-03-28
Add project type handling in cleanappsettings

### LINKED ISSUE #1170: Add project type handling in cleanappsettings
Once we have implemented #1169, we also need to add handling for app settings `project_types` field in the `cleanappsettings` management command.

In short, we should check for existing `AppSetting` objects and delete them if the project type does not match the `project_types` field in the app setting definition.

You may ask, why do we have to add this if we already have the migration set up in #1169? Because sometimes we may make changes in the app settings definitions, e.g. during development or in dev deployments. It's useful to have a simple cleanup command instead of having to add new migrations for every small change.

## Spec

- In the iteration of existing `AppSetting` objects, check for the project type
    * IF `project` is set and the type of the project doesn't match the `project_types` field in the definition for the setting
    * THEN delete that setting
    * AND log an info message for the deletion & its reason
- Do this in the same iteration as the currently existing check (of app setting definition not in plugin), after the current check
    * Can't check for the `project_type` if the definition isn't found :)
- Update command help to convey we perform other checks than for undefined app settings
- Update logging message of existing check to say we deleted it because the definition was not found
- Add tests for project type in `projectroles.tests.test_commands.TestCleanAppSettings`

gromdimon:
## Issue #1170

### DIFF
--- a/projectroles/management/commands/cleanappsettings.py
+++ b/projectroles/management/commands/cleanappsettings.py
@@ -42,10 +42,29 @@ class Command(BaseCommand):
             else:
                 def_kwargs['app_name'] = 'projectroles'
             try:
-                app_settings.get_definition(**def_kwargs)
+                definition = app_settings.get_definition(**def_kwargs)
+                if s.project and s.project.type not in definition.get(
+                    'project_types', ['PROJECT']
+                ):
+                    logger.info(
+                        'Deleting "{}" from project "{}" - project type "{}" does not match allowed types: {}'.format(
+                            get_setting_str(s),
+                            s.project.title,
+                            s.project.type,
+                            definition.get('project_types', []),
+                        )
+                    )
+                    s.delete()
+                elif not definition:
+                    logger.info(
+                        'Deleting "{}" from project "{}" - definition not found'.format(
+                            get_setting_str(s), s.project.title
+                        )
+                    )
+                    s.delete()
             except ValueError:
                 logger.info(
-                    'Deleting "{}" from project "{}"'.format(
+                    'Deleting "{}" from project "{}" - definition not found'.format(
                         get_setting_str(s), s.project.title
                     )
                 )
--- a/projectroles/management/commands/cleanappsettings.py
+++ b/projectroles/management/commands/cleanappsettings.py
@@ -51,7 +51,7 @@ class Command(BaseCommand):
                             get_setting_str(s),
                             s.project.title,
                             s.project.type,
-                            definition.get('project_types', []),
+                            definition.get('project_types', ['PROJECT']),
                         )
                     )
                     s.delete()
--- a/projectroles/management/commands/cleanappsettings.py
+++ b/projectroles/management/commands/cleanappsettings.py
@@ -14,6 +14,8 @@ START_MSG = 'Checking database for undefined app settings..'
 END_MSG = 'OK'
 DEFINITION_NOT_FOUND_MSG = 'definition not found'
 ALLOWED_TYPES_MSG = 'does not match allowed types'
+DELETE_PREFIX_MSG = 'Deleting "{}" from project "{}": '
+DELETE_PROJECT_TYPE_MSG = 'project type "{}" {}: {}'
 
 
 def get_setting_str(db_setting):
@@ -47,11 +49,10 @@ class Command(BaseCommand):
                 definition = app_settings.get_definition(**def_kwargs)
             except ValueError:
                 logger.info(
-                    'Deleting "{}" from project "{}": {}'.format(
-                        get_setting_str(s),
-                        s.project.title,
-                        DEFINITION_NOT_FOUND_MSG,
+                    DELETE_PREFIX_MSG.format(
+                        get_setting_str(s), s.project.title
                     )
+                    + DEFINITION_NOT_FOUND_MSG
                 )
                 s.delete()
                 continue
@@ -59,9 +60,10 @@ class Command(BaseCommand):
                 'project_types', ['PROJECT']
             ):
                 logger.info(
-                    'Deleting "{}" from project "{}": project type "{}" {}: {}'.format(
-                        get_setting_str(s),
-                        s.project.title,
+                    DELETE_PREFIX_MSG.format(
+                        get_setting_str(s), s.project.title
+                    )
+                    + DELETE_PROJECT_TYPE_MSG.format(
                         s.project.type,
                         ALLOWED_TYPES_MSG,
                         definition.get('project_types', ['PROJECT']),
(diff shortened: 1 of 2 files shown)


projectroles/management/commands/cleanappsettings.py
```diff
@@ -42,10 +42,29 @@ def handle(self, *args, **options):
...
+                    logger.info(
+                        'Deleting "{}" from project "{}" - definition not found'.format(
+                            get_setting_str(s), s.project.title
+                        )
+                    )
+                    s.delete()
             except ValueError:
                 logger.info(
-                    'Deleting "{}" from project "{}"'.format(
+                    'Deleting "{}" from project "{}" - definition not found'.format(
```
mikkonie: Here we check for the definition not existing twice, I would like us to remove the repetition. Handle the possibly missing definition first and continue with other checks if it's found.

Also, I don't believe it's possible for `get_definition()` to return `None`? It at least should raise a `ValueError` if not successful. So I'm not sure if we ever hit the condition above.

projectroles/tests/test_commands.py
```diff
@@ -483,6 +483,8 @@ class TestCleanAppSettings(
 ):
     """Tests for cleanappsettings command and associated functions"""
 
+    maxDiff = None
```
mikkonie: Remember to remove this before the final commit for this pr :)

projectroles/management/commands/cleanappsettings.py
```diff
@@ -42,10 +42,29 @@ def handle(self, *args, **options):
             else:
                 def_kwargs['app_name'] = 'projectroles'
             try:
-                app_settings.get_definition(**def_kwargs)
+                definition = app_settings.get_definition(**def_kwargs)
+                if s.project and s.project.type not in definition.get(
+                    'project_types', ['PROJECT']
+                ):
+                    logger.info(
+                        'Deleting "{}" from project "{}" - project type "{}" does not match allowed types: {}'.format(
```
mikkonie: A minor point but the ` - ` syntax is not consistent with SODAR Core logging messages, I recommend using `: ` instead.

projectroles/tests/test_commands.py
```diff
@@ -602,9 +604,19 @@ def test_command_cleanappsettings(self):
...
-                        'from project "{}"'.format(self.project.title)
+                        'from project "{}" - definition not found'.format(
+                            self.project.title
+                        )
+                    ),
+                    # Check for project type
+                    (
+                        CLEAN_LOG_PREFIX
+                        + 'Deleting "settings.example_project_app.project_int_setting" '
+                        'from project "{}" - project type "{}" does not match allowed types: {}'.format(
```
mikkonie: I recommend defining these error strigns as constants in the command module, importing the constants to tests and checking against those instead. Asserting long strings can very easily become annoying when we e.g. want to change the test in code and have to go to tests to modify the strings there.

projectroles/tests/test_commands.py
```diff
@@ -531,7 +533,7 @@ def setUp(self):
         }
         self.setting_int_values = {
             'app_name': EXAMPLE_APP_NAME,
-            'project': self.project,
+            'project': self.category,
```
mikkonie: I recommend splitting the current test function in two, both testing different functionality. One for the undefined setting and one for the setting with a wrong project type. Then, initialize the latter within the tests which asserts it gets deleted. It's good to separate testing specific functionality into separate unit tests, as it makes it easier to spot what exactly goes wrong if and when they fail. Hence the name "unit test" :)

projectroles/tests/test_commands.py
```diff
@@ -39,6 +39,12 @@
 
 EXAMPLE_APP_NAME = 'example_project_app'
 CLEAN_LOG_PREFIX = 'INFO:projectroles.management.commands.cleanappsettings:'
+CLEAN_LOG_DELETE_GHOST = (
+    'Deleting "settings.example_project_app.ghost" from project'
+)
+CLEAN_LOG_DELETE_GHOST_END = 'definition not found'
+CLEAN_LOG_DELETE_INT_SETTING = 'Deleting "settings.example_project_app.project_user_bool_setting" from project'
+CLEAN_LOG_DELETE_INT_SETTING_END = 'does not match allowed types: [\'PROJECT\']'
```
mikkonie: Sorry, I was unclear in my earlier point. Instead of redefining these constants here, we should simply import the constants used in cleanappsettings.py and compare our test output to those, with `{placeholders}` in those constants replaced by the appropriate strings using `format()`.

This way, if we decide to make a change to the wording of these log messages, we don't have to also update strings in the test cases.

projectroles/management/commands/cleanappsettings.py
```diff
@@ -42,11 +44,27 @@ def handle(self, *args, **options):
             else:
                 def_kwargs['app_name'] = 'projectroles'
             try:
-                app_settings.get_definition(**def_kwargs)
+                definition = app_settings.get_definition(**def_kwargs)
             except ValueError:
                 logger.info(
-                    'Deleting "{}" from project "{}"'.format(
-                        get_setting_str(s), s.project.title
+                    'Deleting "{}" from project "{}": {}'.format(
```
mikkonie: Since there is repetition for "Deleting x from project y", I recommend making this prefix a constant with the appropriate `{placeholders}` so we can simply refer and reformat to that, both here and in the test cases.

projectroles/tests/test_commands.py
```diff
@@ -495,6 +501,18 @@ def _make_undefined_setting(self):
         ghost.save()
         return ghost
 
+    def _make_setting_category_type(self):
```
mikkonie: If this is only used once in one test, we probably don't have to add a separate method for it. Instead, it can just be created at the beginning of the test.

projectroles/tests/test_commands.py
```diff
@@ -495,6 +501,18 @@ def _make_undefined_setting(self):
...
+        """Create database setting with category type project"""
+        catsett = AppSetting(
```
mikkonie: Minor point, but note the variable naming. Words should be separated by underscores as per convention :)

projectroles/tests/test_commands.py
```diff
@@ -601,10 +619,45 @@ def test_command_cleanappsettings(self):
...
                     CLEAN_LOG_PREFIX + END_MSG,
                 ],
             )
         self.assertEqual(AppSetting.objects.count(), 5)
+        self.assertQuerysetEqual(
+            AppSetting.objects.filter(name='ghost'),
+            [],
+        )
+
+    def test_command_cleanappsettings_project_types(self):
```
mikkonie: I think we can drop the `cleanappsettings` from the test method names here since we have a separate class for the command anyway. Just call them `test_command_*()`.

(I know, this wasn't originally decided by you but someone else working on this before, but let's clean things up while we're at it :))

PR #1182, opened 2023-03-29
Display app icon for settings in project/user form

### LINKED ISSUE #947: Display app icon for app settings in project/user forms
This would make the origin of these app settings more clear to the user, especially in cases where it e.g. isn't spelled out clearly in the help text.

Instead of `Label or setting name` as the field name, we should have `<i class="iconify" data-icon="plugin.icon"> Label or setting name`.

I believe setting HTML can be easily done by setting the label with `mark_safe()` imported from `django.utils.safestring`.

## Spec

- Add icons to app setting labels in `ProjectForm`
- Add icons to app setting labels in `UserForm`
- Add simple tests to ensure an icon for an app setting is visible in both forms

gromdimon:
## Issue #947

### DIFF
--- a/projectroles/forms.py
+++ b/projectroles/forms.py
@@ -450,7 +450,7 @@ class ProjectForm(SODARModelForm):
             self.fields[s_field].label = format_html(
                 '{} <i class="iconify" title="{}" data-toggle="tooltip" data-icon="{}"></i>',
                 self.fields[s_field].label,
-                plugin.name,
+                plugin.title,
                 plugin.icon,
             )
         else:
--- a/projectroles/forms.py
+++ b/projectroles/forms.py
@@ -90,6 +90,21 @@ class SODARFormMixin:
         self.logger.error(log_msg)
         super().add_error(field, error)  # Call the error method in Django forms
 
+    def get_app_setting_label(self, plugin, label):
+        """Return label for app setting key"""
+        if plugin:
+            return format_html(
+                '{} <i class="iconify" title="{}" data-toggle="tooltip" data-icon="{}"></i>',
+                label,
+                plugin.title,
+                plugin.icon,
+            )
+        else:
+            return format_html(
+                '{} <i class="iconify" title="projectroles" data-icon="mdi-cube"></i>',
+                label,
+            )
+
 
 class SODARForm(SODARFormMixin, forms.Form):
     """Override of Django base form with SODAR Core specific helpers."""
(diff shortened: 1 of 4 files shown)


mikkonie:
This works great and is very well executed! Just a few points:

- Now that I actually see the result, I think it might look better if the icons comes **after** the label, not before. Could we try that? The variable left alignment of the labels can be a bit confusing.
- It would also be helpful if the icon had the app plugin `title` displayed as the `title=""` attribute.
    * If you want, you can also make this fancy by using the Bootstrap 4 tooltip (`data-toggle="tooltip"` etc) here. Just make sure it works properly, as it can have problems with certain parent elements.

projectroles/forms.py
```diff
@@ -444,6 +446,18 @@ def _set_app_setting_notes(self, s_field, s_val):
                 self.fields[
                     s_field
                 ].help_text += ' [Not editable on target sites]'
+        if plugin:
```
mikkonie: Since there's quite a few rows of repetition here, how about creating a `get_app_setting_label()` helper in `SODARModelForm` and calling that from both the project and user forms?

projectroles/forms.py
```diff
@@ -444,6 +446,18 @@ def _set_app_setting_notes(self, s_field, s_val):
...
+            self.fields[s_field].label = format_html(
+                '{} <i class="iconify" title="{}" data-toggle="tooltip" data-icon="{}"></i>',
```
mikkonie: Since the Bootstrap tooltip isn't working here, feel free to remove the `data-toggle` attributes from here and the user form.

projectroles/forms.py
```diff
@@ -89,6 +90,21 @@ def add_error(self, field, error):
         self.logger.error(log_msg)
         super().add_error(field, error)  # Call the error method in Django forms
 
+    def get_app_setting_label(self, plugin, label):
+        """Return label for app setting key"""
+        if plugin:
+            return format_html(
+                '{} <i class="iconify" title="{}" data-toggle="tooltip" data-icon="{}"></i>',
```
mikkonie: There's still a `data-toggle` attribute here, we can remove that.

projectroles/tests/test_ui.py
```diff
@@ -1461,6 +1461,28 @@ def test_settings_fields_category(self):
...
+            self.selenium.find_element(
+                By.CSS_SELECTOR,
+                'div[id="div_id_type"] div select[id="id_type"]',
+            )
+        )
+        select.select_by_value('PROJECT')
+        WebDriverWait(self.selenium, 10)
+        logo = self.selenium.find_element(
+            By.CSS_SELECTOR,
+            'div[id="div_id_settings.example_project_app.project_int_setting"] ',
```
mikkonie: There's a n extra space at the end of this selector, probably not needed?

userprofile/tests/test_ui.py
```diff
@@ -22,3 +24,27 @@ def test_update_button(self):
...
+
+    def test_settings_label_icon(self):
+        """Test existence of settings label icon"""
+        url = reverse('userprofile:settings_update')
+        self.login_and_redirect(
+            self.superuser, url, wait_elem=None, wait_loc='ID'
+        )
+        WebDriverWait(self.selenium, 15)
+
+        logo = self.selenium.find_element(
```
mikkonie: I'm nitpicking here, but technically we're looking for an "icon" here, not a "logo". Might be confusing for someone reading/updating this test case later on :)

PR #1183, opened 2023-03-29
Make timeline extra data only visible to superusers, owners and delegates

### LINKED ISSUE #1171: Make timeline extra data only visible to superusers, owners and delegates
There are some cases in SODAR where a regular user may see some data which is not exactly intended for them via the extra data saved by certain timeline events.

In the future, we might set up some kind of permission levels to events and/or their data. But that's not a priority for current development and needs to be thought out in detail before implementing. Hence, for now, let's change this so that only project owners and delegates are able to access event extra data.

## Spec

- Add new permission `view_extra_data` to `timeline.rules`
    * Only allow that for owners and delegates
- In timeline templates, only display link to extra data
    * IF `project` is set for event AND user has `view_extra_data` perm
    * OR `project` is unset AND user is superuser
- Add the same behaviour in the event details modal
- Update the required permissions for `SiteEventExtraAjaxView` and `EventStatusExtraAjaxView`
- Update tests as needed

gromdimon:
## Issue #1171

### DIFF
--- a/timeline/rules.py
+++ b/timeline/rules.py
@@ -44,3 +44,9 @@ rules.add_perm(
 
 # Allow viewing classified site-wide event
 rules.add_perm('timeline.view_classified_site_event', rules.is_superuser)
+
+# Allow viewing extra data of event
+rules.add_perm(
+    'timeline.view_event_extra_data',
+    pr_rules.is_project_owner | pr_rules.is_project_delegate,
+)
--- a/timeline/rules.py
+++ b/timeline/rules.py
@@ -45,7 +45,7 @@ rules.add_perm(
 # Allow viewing classified site-wide event
 rules.add_perm('timeline.view_classified_site_event', rules.is_superuser)
 
-# Allow viewing extra data of event
+# Allow viewing extra data of project event
 rules.add_perm(
     'timeline.view_event_extra_data',
     pr_rules.is_project_owner | pr_rules.is_project_delegate,
--- a/timeline/static/timeline/js/timeline.js
+++ b/timeline/static/timeline/js/timeline.js
@@ -48,7 +48,7 @@ $(document).ready(function() {
                     'data-url="' + data['status'][i]['extra_status_link'] + '"><i class="iconify" ' +
                     'data-icon="mdi:text-box" title="Status Extra Data" data-toggle="tooltip" data-placement="right">' +
                     '</i></a>';
-                if (data['status'][i]['extra_status_link'] === null || data['status'][i]['extra_status_link'] === '') {
+                if (data['status'][i]['extra_status_link'] === null ) {
                     extraData = '';
                 }
                 tableBody.append($('<tr>')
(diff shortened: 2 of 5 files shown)


mikkonie:
Uh-oh, found a bug!

- Create a `ProjectEvent` with extra data in related `ProjectEventStatus`
- Log in as regular user (non-superuser)
- Navigate to event list view
- Open the details modal from the timestamp link
- Observe that the extra data link is visible :)
    * Clicking on this returns "forbidden", which is good, but we want to hide  the link
- This works on both project and site-wide events

mikkonie:
Still a bit of a bug:

- Project contributor now can't see links to extra data, that's good
- Project owner and delegate see link (good) but when clicked it returns as "forbidden" (bad)

mikkonie:
Sadly the bug I reported yesterday is still there: project owner and delegate are unable to access `ProjectEventStatus` extra data. The link is visible in the modal, but the request returns "forbidden".

mikkonie:
Bug time! :)

- Create an event without a project, add some extra data to it
- Log in as superuser
- Navigate to "site-wide events" or "all timeline events"
- Click on the extra data link
- Observe `Error: Not Found` message in modal
    * (If you add extra data to the same even'ts status, it works fine)

timeline/rules.py
```diff
@@ -44,3 +44,9 @@
 
 # Allow viewing classified site-wide event
 rules.add_perm('timeline.view_classified_site_event', rules.is_superuser)
+
+# Allow viewing extra data of event
```
mikkonie: Just a small note, maybe make it clear in this comment that this perm is specifically for viewing project events.

timeline/static/timeline/js/timeline.js
```diff
@@ -48,7 +48,7 @@ $(document).ready(function() {
                     'data-url="' + data['status'][i]['extra_status_link'] + '"><i class="iconify" ' +
                     'data-icon="mdi:text-box" title="Status Extra Data" data-toggle="tooltip" data-placement="right">' +
                     '</i></a>';
-                if (data['status'][i]['extra_status_link'] === null) {
+                if (data['status'][i]['extra_status_link'] === null ) {
```
mikkonie: Extra space here after `null` :)

timeline/templates/timeline/_list_item.html
```diff
@@ -37,12 +39,22 @@
...
-      </a>
+      {% if event.project and can_view_extra_data %}
+        <a class="sodar-tl-link-extra-data text-primary pull-right"
+           data-url="{% url 'timeline:ajax_extra_project' projectevent=event.sodar_uuid %}">
+          <i class="iconify" data-icon="mdi:text-box" title="Extra Data"
+             data-toggle="tooltip" data-placement="right">
+          </i>
+        </a>
+      {% endif %}
+      {% if not event.project and user.is_superuser %}
```
mikkonie: This could also be an `{% elif %}` to get rid of one extra template row :)

Issue #1129, opened 2023-02-10
Add PROJECT_USER app setting cleanup to cleanappsettings


mikkonie:
The logic for cleaning up `PROJECT-USER` app settings upon role deletion in #1128 also needs to be implemented in the `cleanappsettings` management command.

## Spec

- FOR each app setting with `PROJECT-USER` scope
- IF user has no role in project (including inherited)
- THEN delete setting

mikkonie:
Done by @gromdimon.

PR #1186, opened 2023-03-31
Projectroles/site scope settings

### LINKED ISSUE #1184: Add SITE scope to app settings
Several upcoming planned features, e.g. admin actions such as #24 and #1013, require us to have some type of site-wide settings which can be mutable during runtime. Think of Django settings but changeable while the server is running.

Since we already have the awesome app settings framework, let's use it for this. In addition to `PROJECT`, `USER` and `PROJECT-USER`, this would simply be a new scope for the app settings: settings not linked to any project or user.

## Spec

- Add new `SITE` scope to app settings
    * Should work as existing app settings, except not tied to any project or user
    * Only once instance of this type of setting (by app name and setting name) should exist on a site (the `unique_together` constraint in the model should already take care of this)
- Update `AppSettingAPI` and validation to accept settings of this scope
    * For a `SITE` setting, `user` and `project` must be `None`
    * `project_type` is disregarded, of course
- Add example setting in `example_project_app`
- Add tests

gromdimon:
## Issue #1184


gromdimon:
### Implementing #1129 in this PR as well

projectroles/app_settings.py
```diff
@@ -42,7 +44,7 @@
     #: Example ::
     #:
     #:     'example_setting': {
-    #:         'scope': 'PROJECT',  # PROJECT/USER
+    #:         'scope': 'PROJECT',  # PROJECT/USER/SITE
```
mikkonie: Just noticed that `PROJECT-USER` is missing from this example. Not your fault as it should have been there before, but let's add it :)

projectroles/models.py
```diff
@@ -48,6 +49,7 @@
     ('JSON', 'Json'),
 ]
 APP_SETTING_VAL_MAXLENGTH = 255
+APP_SETTING_SCOPE_SITE = SODAR_CONSTANTS['APP_SETTING_SCOPE_SITE']
```
mikkonie: Minor note, but please move this a few lines up under "SODAR constants" :)

projectroles/models.py
```diff
@@ -820,6 +820,24 @@ def get_setting_value(
         if not app_name == 'projectroles':
             query_parameters['app_plugin__name'] = app_name
         setting = super().get_queryset().get(**query_parameters)
+        # plugin = get_app_plugin(app_name)
```
mikkonie: The validation is already done when we call this in `AppSettingAPI`, so no further checks should be needed here? Removing the check for project and user above is correct.

PR #1192, opened 2023-04-03
PROJECT-USER scope handling

### LINKED ISSUE #1128: Add PROJECT_USER app setting handling on RoleAssignment delete
When looking into #321 with @gromdimon, I noticed we are not properly cleaning up `PROJECT-USER` scope app settings upon `RoleAssignment` deletion in in `delete_assignment()`.

Furthermore, I realized the logic for this will change with #638. Thus, the spec becomes as follows:

## Spec

- IF a `RoleAssignment` is deleted
    * AND user has no inherited role
    * AND project type is `CATEGORY` AND user has no role in child categories/projects
- THEN delete all of this user's `PROJECT-USER` scope app settings for the current project/category
    * AND its child categories/projects
- Add/update tests to ensure this deletion takes place correctly
    * Remember to add different tests for no inheritance / inheritance / child categories/projects

Needless to say, this requires first implementing and merging #638.

**Hint:** To work with inherited roles, you generally don't want to be doing direct `RoleAssignment` queries. Instead, look into the `Project` model API methods such as `get_role()` and `get_roles()`, which can be used to grab both inherited and local roles.

### LINKED ISSUE #1129: Add PROJECT_USER app setting cleanup to cleanappsettings

gromdimon:
## Issues: #1128 and #1129 

However, I found a more straightforward solution for issues and implemented it. If it's not suitable or outperforming worse, let me know and I will refactor it :)


projectroles/app_settings.py
```diff
@@ -626,6 +626,44 @@ def delete(cls, app_name, setting_name, project=None, user=None):
             )
         )
 
+    @classmethod
+    def delete_settings_by_scope(
```
mikkonie: This is a good idea. I would like the following minor changes:

- Rename method into `delete_by_scope()`
- Note in the docstring that this deletes all settings within a scope for a specific project/user combination
- Change signature so that `scope` comes before `project` and `user`
- Remove the `=None` from `scope` as it must always be set
- Add `SITE` scope support perhaps? :)

projectroles/management/commands/cleanappsettings.py
```diff
@@ -16,6 +16,7 @@
 ALLOWED_TYPES_MSG = 'does not match allowed types'
 DELETE_PREFIX_MSG = 'Deleting "{}" from project "{}": '
 DELETE_PROJECT_TYPE_MSG = 'project type "{}" {}: {}'
+DELETE_SCOPE_MSG = 'user {} has no Role in the project'
```
mikkonie: For logging conventions we can simplify this into `user {} has no role in project`.

projectroles/management/commands/cleanappsettings.py
```diff
@@ -70,4 +71,17 @@ def handle(self, *args, **options):
                     )
                 )
                 s.delete()
+
+        db_settings = AppSetting.objects.filter(
+            project__isnull=False, user__isnull=False
+        )
+        for s in db_settings:
+            if s.project.get_role(s.user, inherited_only=True) is None:
```
mikkonie: Here we have a problem. The original spec says: "IF user has no role in project (including inherited)". Here, we set `inherited_only=True` so we only delete settings if there are inherited roles in addition to a local one.

Removing the `inherited_only` arg should fix this.

As a very small note, we can just use `not` instead of `is None` for brevity :)

projectroles/views.py
```diff
@@ -1757,8 +1760,18 @@ def delete_assignment(self, request, instance):
...
         # Delete object itself
         instance.delete()
-        # Remove project star from user if it exists
-        app_settings.delete('projectroles', 'project_star', project, user)
+
+        # Delete corresponding PROJECT_USER settings
+        if (
+            not project.get_role(user, inherited_only=True)
+            and project.type == PROJECT_TYPE_CATEGORY
+            and not RoleAssignment.objects.filter(
```
mikkonie, line 1768: Here we also have an issue. This query looks for only the immediate child project, but the category structure can be deep and a category may have multiple levels of children under it. You should use `Project.get_children()` instead.

PR #1193, opened 2023-04-04
1190 refactor and cleanup appsettingapi

### LINKED ISSUE #1190: Refactor and cleanup AppSettingAPI
I have noticed some things to refactor and clean up in `AppSettingAPI`.

NOTE: These are based on me looking at the module and I could be incorrect about some details, so ask me if anything seems unclear. Also, feel free to let me know if you notice any other similar things worth refactoring.

## Spec

- Simplify `_check_project_and_user()`
    * This is getting too spaghetti-like. Instead, set up a dict in module constants with valid project/user scopes and validate against that
    * See example in comments below
- Move `_log_debug()` from inside `set()` into a `_log_set_debug()` classmethod
    * This makes `set()` look less messy
- Remove duplicate setting definition retrieval in `set()`
    * We grab the definition at the start of the method with `get_definition()`..
    * ..However, under `except AppSetting.DoesNotExist`, we retrieve the setting def again into `s_def`.
        + We already have the definition so this is redundant
        + We do still need to get the app plugin and model, so make sure to leave those in
- Improve scope and project/user checks  in `set()`
    * Move calls to `_check_scope()` and `_check_project_and_user()` right after `get_definition()`
    * Remove duplicate manual checks for project and user args
- Replace manual project and user checks in `delete()`
    * Use `_check_project_and_user()` instead
    * This will also allow us to properly check for arguments for deleting a `SITE` scope setting..
- Make sure tests catch all these changes and assert expected results

gromdimon:
## Issue #1190

### DIFF
--- a/projectroles/app_settings.py
+++ b/projectroles/app_settings.py
@@ -36,6 +36,12 @@ APP_SETTING_DEFAULT_VALUES = {
     'BOOLEAN': False,
     'JSON': {},
 }
+APP_SETTING_VALID_SCOPE_VALIDATION = {
+    APP_SETTING_SCOPE_PROJECT: {'project': True, 'user': False},
+    APP_SETTING_SCOPE_USER: {'project': False, 'user': True},
+    APP_SETTING_SCOPE_PROJECT_USER: {'project': True, 'user': True},
+    APP_SETTING_SCOPE_SITE: {'project': False, 'user': False},
+}
 
 # Define App Settings for projectroles app
 PROJECTROLES_APP_SETTINGS = {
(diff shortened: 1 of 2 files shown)


projectroles/app_settings.py
```diff
@@ -36,6 +36,12 @@
     'BOOLEAN': False,
     'JSON': {},
 }
+APP_SETTING_VALID_SCOPE_VALIDATION = {
```
mikkonie, line 39: I prefer `APP_SETTING_SCOPE_ARGS` here :)

projectroles/app_settings.py
```diff
@@ -101,37 +107,35 @@ class AppSettingAPI:
...
-                    'User unset for setting with project_user scope'
-                )
-        elif scope == APP_SETTING_SCOPE_SITE:
-            if project:
-                raise ValueError('Project set for setting with site scope')
-            if user:
-                raise ValueError('User set for setting with site scope')
+        :raise: ValueError if parameters are not valid for scope
+        """
+        valid_scope = APP_SETTING_VALID_SCOPE_VALIDATION.get(scope)
```
mikkonie, line 117: My original idea was to simplify this even further. We can just do:

```
if not APP_SETTING_SCOPE_ARGS[scope] == {'project': project is not None, 'user': user is not None}:
    raise ValueError()  # etc etc
```

Also, you should rebase to dev and call this also in your newly created `delete_by_scope()` method :)

projectroles/app_settings.py
```diff
@@ -480,37 +513,8 @@ def set(
...
-        ):
-            raise ValueError('Project and user are both unset')
-        elif (
-            setting_def.get('scope') == APP_SETTING_SCOPE_SITE
-            and project
-            or setting_def.get('scope') == APP_SETTING_SCOPE_SITE
-            and user
-        ):
-            raise ValueError('Project or user is set for site scope setting')
+        cls._check_project_and_user(setting_def.get('scope'), project, user)
```
mikkonie, line 517: We should also do `_check_scope()` here.

projectroles/app_settings.py
```diff
@@ -570,16 +576,14 @@ def _log_debug(action, app_name, setting_name, value, project, user):
...
-                bool(s_def['user_modifiable'])
-                if 'user_modifiable' in s_def
+                bool(setting_def['user_modifiable'])
+                if 'user_modifiable' in setting_def
                 else True
             )
 
-            cls._check_scope(s_def['scope'])
-            cls._check_project_and_user(s_def['scope'], project, user)
+            cls._check_project_and_user(setting_def['scope'], project, user)
```
mikkonie, line 586: No need to check project and user again here, as already we do it earlier.

PR #1195, opened 2023-04-04
Custom template/include path setting

### LINKED ISSUE #1049: Enable setting custom include template path
Currently, custom template includes are hardcoded to `{SITE}/templates/include/`. However, in case of dynamic includes for e.g. different site instances, we want to have these outside of the repository.

Hence, we need to add a Django setting `PROJECTROLES_TEMPLATE_INCLUDE_PATH` defaulting to the aforementioned path and read the include path from this variable.

Obviously, this should be treated as a full path rather than a relative one to allow for includes from outside of the repo.

gromdimon:
## Issue #1049


mikkonie:
Some notes:

- Actual use of the path in templates is missing.
    * Hint: See `grep -nir --exclude-dir "docs" --include "*.html" "include/"`
- The default value should be `os.path.join(APPS_DIR, 'templates', 'include')`

PR #1197, opened 2023-04-04
Projectroles/app settings api refactoring

gromdimon:
### Small note: 
As I found now, previously it was possible to delete multiple settings with one `app_settings_api.delete()` command. For that behaviour you just had to NOT provide project and/or user arguments. However, with new refactoring such behaviour is being caught in `_check_project_and_user()` method, and as a result it causes `ValueError`. I changed 2 tests, that were falling due to new validation step, but if we would like to move back to previous deleting realisation, I have to change it as well. 

The failing tests: `test_delete_scope_user_params_none (projectroles.tests.test_app_settings.TestAppSettingAPI)` and `test_delete_scope_project_params_none (projectroles.tests.test_app_settings.TestAppSettingAPI)`

### DIFF
--- a/projectroles/app_settings.py
+++ b/projectroles/app_settings.py
@@ -36,6 +36,12 @@ APP_SETTING_DEFAULT_VALUES = {
     'BOOLEAN': False,
     'JSON': {},
 }
+APP_SETTING_SCOPE_ARGS = {
+    APP_SETTING_SCOPE_PROJECT: {'project': True, 'user': False},
+    APP_SETTING_SCOPE_USER: {'project': False, 'user': True},
+    APP_SETTING_SCOPE_PROJECT_USER: {'project': True, 'user': True},
+    APP_SETTING_SCOPE_SITE: {'project': False, 'user': False},
+}
 
 # Define App Settings for projectroles app
 PROJECTROLES_APP_SETTINGS = {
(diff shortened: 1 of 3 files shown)


projectroles/app_settings.py
```diff
@@ -101,37 +107,22 @@ class AppSettingAPI:
...
-                )
-            if not user:
-                raise ValueError(
-                    'User unset for setting with project_user scope'
+        if not APP_SETTING_SCOPE_ARGS[scope] == {
+            'project': project is not None,
+            'user': user is not None,
+        }:
+            raise ValueError(
+                'Project and/or user are set incorrect for setting with {} scope'.format(
```
mikkonie: I believe this row goes beyond 80 characters in length :) (black doesn't catch ones with strings)

projectroles/app_settings.py
```diff
@@ -278,6 +269,34 @@ def _compare_value(cls, setting_obj, input_value):
             return bool(int(setting_obj.value)) == input_value
         return setting_obj.value == str(input_value)
 
+    @classmethod
+    def _log_set_debug(
+        cls, action, app_name, setting_name, value, project, user
+    ):
+        """
+        Helper method for logging setting changes in set() method.
```
mikkonie, line 276: Missing empty line between doctsring description and params :)

mikkonie:
This pr seems fine in its current state, good work. I'll wait to merge this after the others if there are indeed some conflicts as we discussed.

PR #1199, opened 2023-04-05
Projectroles/site wide timeline events for remote site operations

### LINKED ISSUE #746: Add site-wide timeline events for remote site operations
We should log these site-wide events for admins, as right now no trace is saved from these actions. That makes debugging difficult when we have some errors regarding this.

## Spec

- Create timeline events for following actions:
    * `remote_site_create`
        + Add in `RemoteSiteModifyMixin`
        + As description, add: `create remote site {{{ remote_site }}}`
    * `remote_site_update`
        + Add in `RemoteSiteModifyMixin` (should be a modification of the previous call, of course)
        + As description, add: `update remote site {{{ remote_site }}}`
    * `remote_site_delete`
        + Add in `RemoteSiteDeleteView`
        + As description, add: `delete remote site {{{ remote_site }}}`
    * `remote_project_update`
        + Add in `RemoteProjectBatchUpdateView`
        + To be created in POST once confirmed
        + As extra data, add list of access changes (see `modifying_access`) 
    * `remote_project_sync`
        + Add in `RemoteProjectSyncView`
        + NOTE: we only want to log UI-based manual syncs here, so not the management command or the upcoming periodic sync via celery..
- For each event: set `classified=True`
- Add checks for timeline event creation for each of these actions under `TestRemoteSiteCreateView` etc.

gromdimon:
## Issue #746

### DIFF
--- a/projectroles/tests/test_views.py
+++ b/projectroles/tests/test_views.py
@@ -4847,6 +4847,7 @@ class TestRemoteSiteCreateView(RemoteSiteMixin, TestViewsBase):
 
     def test_create_target(self):
         """Test creating a target site"""
+        timeline = get_backend_api('timeline_backend')
         self.assertEqual(RemoteSite.objects.all().count(), 0)
         values = {
             'name': REMOTE_SITE_NAME,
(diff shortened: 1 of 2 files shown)


gromdimon:
While writing test I found out, that `TestRemoteProjectSyncView` was not implemented at all. The reason for that, as I understand, could be difficulty, that you need to run simultaniously two servers while testing functionality. I wrote an example test, that could theoratically work, but I still didn't find a solution for emulating 2 servers...

From that point I have a question: What should be done with that?
Maybe use some 'mocking' techniques to emulate a source server responces? Or implement another mixin, that somehow sends responses? I have not so much experience with such staff, so will be happy for any comments!

PR #1200, opened 2023-04-06
Projectroles/periodic remote project sync

### LINKED ISSUE #813: Add periodic remote project sync
Currently, projects intended to be deployed as target sites must themselves set up periodic remote sync. We should, instead, set up a periodic Celery job to run the sync on a set interval.

Requires #1198.

### Settings and scaffolding

- Add Django setting `PROJECTROLES_TARGET_SYNC_ENABLE`
    * Boolean type, retrieve from env var with the same name, default = False
- Add Django setting `PROJECTROLES_TARGET_SYNC_INTERVAL`
    * Integer type, get from env, default=5 (this will be in minutes)
- Those settings are considered **optional**

### Refactor sync logic for reuse

- Move sync data retrieval functionality from `RemoteProjectSyncView.get()` into a method in `RemoteProjectAPI`, call it e.g. `get_remote_data()`
    * You guessed it: in the future this functionality will be called both from the view and from the celery task :)
    * The `context` stuff is for the UI of course, we won't need that in the reusable method. You'll probably figure out what else to reuse and what is there just for the UI
    * Call the new method in view, pass data to `sync_remote_data()`

### Set up Tasks

- Add `tasks.py` under `projectroles`
    *  Hint: For an example in how to set a periodic task, see e.g. `samplesheets.tasks_celery` in SODAR
- Define remote project sync task    
    * Init `RemoteProjectAPI`
    * Get `RemoteSite`of type `SOURCE`
        + There can be 0 or 1 of these
    * If site exists, call the sync data retrieval method moved to the mixin earlier
    * If data is received, call `sync_remote_data()` and pass that data onwards
    * This should sync data on the target site!
- Set up periodic execution of task
    * See `setup_periodic_tasks()` in `samplesheets.tasks_celery`.. Well, basically just copy it ;)
    * Add this task:
        + IF `settings.PROJECTROLES_SITE_MODE==SITE_MODE_TARGET`
        + AND `getattr(settings, 'PROJECTROLES_TARGET_SYNC_ENABLE', False)`
    * Set interval as `getattr(settings, 'PROJECTROLES_TARGET_SYNC_INTERVAL', 5) * 60`
        + Setting is in minutes, value here will be in seconds..

### Tests, docs etc.

- Add tests for the task method
    * Test with different params to ensure data is changed and there are no horrible crashes
    * No need to test  the full changed data, this should be identical to the view which is already tested
- Add settings to docs
- Add note on remote project sync about periodic sync if enabled and Celery process is running

### LINKED ISSUE #1198: Add celery setup
Required for e.g. #813. While setting up the celery workers is, in the end, the job of a site using SODAR Core, we should also have an example and dev implementation within this project for running the celery and celerybeat process.

## Spec

- Add `redis>=4.4.4` dependency to `requirements/base.txt`
- Add `celery` target to `Makefile` (see below)
- Add note for celery process in documentation
    * Where ever `make serve` is mentioned, note that to enable periodic tasks you also need to run `make celery`.

Celery target for Makefile (see SODAR Makefile for an example):

```
.PHONY: celery
celery:
    celery -A config worker -l info --beat
```

I don't *think* this should require any further changes, but feel free to verify and compare with the redis settings in SODAR.

gromdimon:
### TODO:
- [x] Set up Celery
- [x] Document celery
- [x] Add django settings
- [x] Add `RemoteProjectAPI.get_remote_data()`
- [x] Set up tasks
- [ ] Tests
- [x] Docs


gromdimon:
### TODO:
- [ ] Document redis-server [start](https://stackoverflow.com/questions/36088409/error-111-connecting-to-localhost6379-connection-refused-django-heroku)

PR #1205, opened 2023-04-14
Site-wide tl events for remote-site changes

gromdimon:
Previous PR: #1199

### DIFF
--- a/projectroles/management/commands/addremotesite.py
+++ b/projectroles/management/commands/addremotesite.py
@@ -7,6 +7,7 @@ from django.db import transaction
 
 from projectroles.management.logging import ManagementCommandLogger
 from projectroles.models import RemoteSite, SODAR_CONSTANTS
+from projectroles.plugins import get_backend_api
 
 
 User = auth.get_user_model()
@@ -140,6 +141,16 @@ class Command(BaseCommand):
                 'user_display': user_diplsay,
             }
             site = RemoteSite.objects.create(**create_values)
+
+        timeline = get_backend_api('timeline_backend')
+        tl_event = timeline.add_event(
+            project=None,
+            app_name='projectroles',
+            event_name='remote_site_create',
+            description=description,
+            classified=True,
+        )
+        tl_event.set_status('OK')
 
         logger.info(
             'Created remote site "{}" with mode {}'.format(site.name, site.mode)
--- a/projectroles/management/commands/addremotesite.py
+++ b/projectroles/management/commands/addremotesite.py
@@ -143,14 +143,14 @@ class Command(BaseCommand):
             site = RemoteSite.objects.create(**create_values)
 
         timeline = get_backend_api('timeline_backend')
-        tl_event = timeline.add_event(
+        timeline.add_event(
             project=None,
             app_name='projectroles',
             event_name='remote_site_create',
             description=description,
             classified=True,
+            status_type='OK',
         )
-        tl_event.set_status('OK')
 
         logger.info(
             'Created remote site "{}" with mode {}'.format(site.name, site.mode)
--- a/projectroles/tests/test_views.py
+++ b/projectroles/tests/test_views.py
@@ -4986,6 +4986,10 @@ class TestRemoteSiteCreateView(RemoteSiteMixin, TestViewsBase):
     def test_create_target(self):
         """Test creating a target site"""
         timeline = get_backend_api('timeline_backend')
+        tl_event = ProjectEvent.objects.filter(
+            event_name='target_site_create'
+        ).first()
+        self.assertIsNone(tl_event)
         self.assertEqual(RemoteSite.objects.all().count(), 0)
         values = {
             'name': REMOTE_SITE_NAME,
@@ -5027,6 +5031,10 @@ class TestRemoteSiteCreateView(RemoteSiteMixin, TestViewsBase):
     def test_create_source(self):
         """Test creating a source site as target"""
         timeline = get_backend_api('timeline_backend')
+        tl_event = ProjectEvent.objects.filter(
+            event_name='source_site_set'
+        ).first()
+        self.assertIsNone(tl_event)
         self.assertEqual(RemoteSite.objects.all().count(), 0)
         values = {
             'name': REMOTE_SITE_NAME,
@@ -5137,6 +5145,10 @@ class TestRemoteSiteUpdateView(RemoteSiteMixin, TestViewsBase):
     def test_update(self):
         """Test updating target site as source"""
         timeline = get_backend_api('timeline_backend')
+        tl_event = ProjectEvent.objects.filter(
+            event_name='target_site_update'
+        ).first()
+        self.assertIsNone(tl_event)
         self.assertEqual(RemoteSite.objects.all().count(), 1)
         values = {
             'name': REMOTE_SITE_NEW_NAME,
@@ -5247,6 +5259,10 @@ class TestRemoteSiteDeleteView(RemoteSiteMixin, TestViewsBase):
     def test_delete(self):
         """Test deleting the remote site"""
         timeline = get_backend_api('timeline_backend')
+        tl_event = ProjectEvent.objects.filter(
+            event_name='target_site_delete'
+        ).first()
+        self.assertIsNone(tl_event)
         self.assertEqual(RemoteSite.objects.all().count(), 1)
         with self.login(self.user):
             response = self.client.post(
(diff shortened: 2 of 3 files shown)


projectroles/tests/test_views.py
```diff
@@ -5012,12 +5013,20 @@ def test_create_target(self):
         }
         model_dict = model_to_dict(site)
         self.assertEqual(model_dict, expected)
+
+        tl_event = (
+            timeline.get_project_events(project=None, classified=True)
+            .order_by('-pk')
+            .first()
+        )
+        self.assertEqual(tl_event.event_name, 'target_site_create')
```
mikkonie, line 5025: I suggest updating the tests as follows:

- Before posting the creation, query for events with the specific name (e.g. `target_site_create`) and assert none of them exist
- After creation, query for events with the name again and assert one is returned

This would be consistent with existing tests and also asserts that the specific action is responsible for the event creation.

projectroles/tests/test_views.py
```diff
@@ -5364,6 +5421,83 @@ def test_post_update(self):
...
 
+        tl_event = (
+            timeline.get_project_events(project=None, classified=True)
+            .order_by('-pk')
+            .first()
+        )
+        self.assertEqual(tl_event.event_name, 'batch_update_remote')
+
+
+# class TestRemoteProjectSyncView(
```
mikkonie: This probably shouldn't be commented out?

projectroles/views.py
```diff
@@ -2893,6 +2893,13 @@ def form_valid(self, form):
         else:
             form_action = 'created'
         self.object = form.save()
+
+        # Create timeline event
+        timeline = get_backend_api('timeline_backend')
+        self.create_timeline_event(
```
mikkonie: The `if timeline` condition should be moved here from the `create_timeline_event()` method so we only call it if timeline is enabled on the SODAR Core implementing site.

projectroles/views.py
```diff
@@ -2901,6 +2908,27 @@ def form_valid(self, form):
...
+        else:
+            event_name = 'target_site_{}'.format(status)
+        if timeline:
+            timeline.add_event(
+                project=None,
+                app_name=APP_NAME,
+                user=user,
+                event_name=event_name,
+                description='{} remote site "{}"'.format(
+                    status, remote_site.name
```
mikkonie: We should add the form fields (name, url, description, user display and secret) as extra data to the event.

projectroles/views.py
```diff
@@ -2901,6 +2908,27 @@ def form_valid(self, form):
...
+            event_name = 'source_site_{}'.format(status)
+        else:
+            event_name = 'target_site_{}'.format(status)
+        if timeline:
+            timeline.add_event(
+                project=None,
+                app_name=APP_NAME,
+                user=user,
+                event_name=event_name,
+                description='{} remote site "{}"'.format(
```
mikkonie: The description could also contain whether the created site was a source or a target site.

projectroles/views.py
```diff
@@ -2901,6 +2908,27 @@ def form_valid(self, form):
...
+    ):
+        """Create timeline event for remote site creation/update"""
+        status = form_action if form_action == 'set' else form_action[0:-1]
+        if remote_site.mode == SITE_MODE_SOURCE:
+            event_name = 'source_site_{}'.format(status)
+        else:
+            event_name = 'target_site_{}'.format(status)
+        if timeline:
+            timeline.add_event(
+                project=None,
```
mikkonie: Setting the event status is missing, hence these events appear with the `INIT` status in timeline. You should set the `OK` status when creating the event.

projectroles/views.py
```diff
@@ -2964,6 +2992,23 @@ class RemoteSiteDeleteView(
...
+        """Override get_success_url() to add message"""
+        timeline = get_backend_api('timeline_backend')
+        if timeline:
+            event_name = (
+                'source_site_delete'
+                if self.object.mode == SITE_MODE_SOURCE
+                else 'target_site_delete'
+            )
+            description = 'delete remote site "{}"'.format(self.object.name)
+            timeline.add_event(
```
mikkonie: Status also missing from here, so we get events with a perpetual `INIT` status. Also check other added events for the same issue.

projectroles/views.py
```diff
@@ -2964,6 +2992,23 @@ class RemoteSiteDeleteView(
     slug_field = 'sodar_uuid'
 
     def get_success_url(self):
+        """Override get_success_url() to add message"""
+        timeline = get_backend_api('timeline_backend')
+        if timeline:
+            event_name = (
+                'source_site_delete'
```
mikkonie: We could simply modify the prefix here instead of the entire event name.

mikkonie:
One further thing I failed to mention in the original spec: Creating a remote site using the `addremotesite` management command should also create a similar timeline event.

projectroles/tests/test_views.py
gromdimon: Oh, that was a class, that has not to be tested (as 2 servers in run are needed for that)

projectroles/tests/test_views.py
```diff
@@ -5012,12 +5016,21 @@ def test_create_target(self):
...
         with self.login(self.user):
             self.assertRedirects(response, reverse('projectroles:remote_sites'))
 
     @override_settings(PROJECTROLES_SITE_MODE=SITE_MODE_TARGET)
     def test_create_source(self):
         """Test creating a source site as target"""
+        tl_event = ProjectEvent.objects.filter(
+            event_name='source_site_set'
+        ).first()
+        self.assertIsNone(tl_event)
```
mikkonie: You can simplify this further by using `count()` or `exists()` and assert true/false.

projectroles/views.py
```diff
@@ -2901,6 +2909,36 @@ def form_valid(self, form):
...
+            extra_data={
+                'name': remote_site.name,
+                'url': remote_site.url,
+                'description': remote_site.description,
+                'mode': remote_site.mode,
+                'user display': remote_site.user_display,
+                'secret': remote_site.secret,
+            },
+        )
+        tl_event.set_status('OK')
```
mikkonie: Hint: you can simplify this by setting `status_type='OK'` in `add_event()` :)

projectroles/views.py
```diff
@@ -2901,6 +2909,36 @@ def form_valid(self, form):
...
+        else:
+            event_name = 'target_site_{}'.format(status)
+
+        tl_event = timeline.add_event(
+            project=None,
+            app_name=APP_NAME,
+            user=user,
+            event_name=event_name,
+            description='{} remote {} site "{}"'.format(
+                status, remote_site.mode, remote_site.name
```
mikkonie: I think `remote_site.mode` looks better in the description if set to lowercase with `lower()`.

projectroles/management/commands/addremotesite.py
```diff
@@ -141,6 +142,16 @@ def handle(self, *args, **options):
...
 
+        timeline = get_backend_api('timeline_backend')
+        tl_event = timeline.add_event(
+            project=None,
+            app_name='projectroles',
+            event_name='remote_site_create',
+            description=description,
+            classified=True,
+        )
+        tl_event.set_status('OK')
```
mikkonie: Here we can still move the status setting in `add_event()`.

projectroles/views.py
```diff
@@ -2901,6 +2909,36 @@ def form_valid(self, form):
...
+            event_name = 'source_site_{}'.format(status)
+        else:
+            event_name = 'target_site_{}'.format(status)
+
+        timeline.add_event(
+            project=None,
+            app_name=APP_NAME,
+            user=user,
+            event_name=event_name,
+            description='{} remote {} site "{}"'.format(
```
mikkonie: One thing I missed in my earlier review (sorry about that): it would be great if we can have a reference object for the remote site for each of these events. That way we can get a history of all the actions done to each remote site in the timeline UI.

So, instead of just having the site name there, use a label in the description and `add_object()` to add a reference to the `RemoteSite` object. Something like this:
```
# ...
description='bla bla {remote_site} bla bla'
# ...
tl_event.add_object(obj=remote_site, label='remote_site', name=remote_site.name)
```

`TimelineAPI._get_ref_description()` already supports `RemoteSite`, so no changes should be needed there.

Again, sorry for missing this earlier, but I think this just adds a small bit of implementation to your pr and doesn't require changing much. Otherwise this pr is looking great!

gromdimon: It's not a problem at all. However, I found out, that previously we used `tl_desc = 'bla bla {{{}}} bla bla'.format('remote_site')`. And that's why I wrote similarly this objects references. 

If I should reform it to `'bla bla {{remote_site}} bla'` without `.format()`, let me know!

PR #1235, opened 2023-06-15
Add url changing in `assert_response*` methods

gromdimon:
### Notes:

The current implementation looks way too user-unfriendly as the cleanup_method has to have self.dynamic_url = url (where url is newly formed url). Besides the self.dynamic_url has to be protected from overwriting.

Bute there are also good points as well:
 - Nothing crashes after changings
 - It is internal function, that is used only in tests (so developers would be able to find the way of implementation the proper cleanup_methods)
 
So here is an example of usage the new dynamic_url (from SODAR):
```
    def _update_request_url(self):
        """Cleanup method"""
        self.irods_request = self.create_request()
        self.url = reverse(
            'samplesheets:api_irods_request_accept',
            kwargs={'irodsdatarequest': self.irods_request.sodar_uuid},
        )  # Update the instance variable

# Test
            irods_request = self.create_request()
            url_accept = reverse(
                'samplesheets:api_irods_request_accept',
                kwargs={'irodsdatarequest': irods_request.sodar_uuid},
            )
            self.assert_response_api(
                url_accept,
                user,
                200,
                method='POST',
                data={'confirm': True},
                cleanup_method=_update_request_url,
            )
```

### DIFF
--- a/projectroles/tests/test_permissions.py
+++ b/projectroles/tests/test_permissions.py
@@ -123,6 +123,10 @@ class TestPermissionBase(TestPermissionMixin, TestCase):
     NOTE: To use with DRF API views, you need to use APITestCase
     """
 
+    def setUp(self):
+        super().setUp()
+        self.url = None
+
 
 class TestProjectPermissionBase(
     ProjectMixin,
@@ -157,6 +161,7 @@ class TestProjectPermissionBase(
         )
 
     def setUp(self):
+        super().setUp()
         # Init roles
         self.init_roles()
         # Init users
--- a/projectroles/tests/test_permissions.py
+++ b/projectroles/tests/test_permissions.py
@@ -40,11 +40,6 @@ REMOTE_SITE_SECRET = build_secret()
 class TestPermissionMixin:
     """Helper class for permission tests"""
 
-    # Special attribute for dynamic URL changing
-    # for use-cases where the URL depends on the object
-    # that is being changed in cleanup_method.
-    dynamic_url = None
-
     def send_request(self, url, method, req_kwargs):
         req_method = getattr(self.client, method.lower(), None)
         if not req_method:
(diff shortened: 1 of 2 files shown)


projectroles/tests/test_permissions.py
```diff
@@ -40,6 +40,11 @@
 class TestPermissionMixin:
     """Helper class for permission tests"""
 
+    # Special attribute for dynamic URL changing
+    # for use-cases where the URL depends on the object
+    # that is being changed in cleanup_method.
+    dynamic_url = None
```
mikkonie: Couldn't we just do the following:

- Update an URL variable (e.g. `self.url`) in `cleanup_method`
- Pass that URL to the `assert_*()` method(s) as the `url` argument, whenever we call the method in a loop

That would be more clear, as we would not be dependent on the existence of a specifically named class member. variable Looking at the implementation, this should work just fine, as we now disregard the `url` argument completely if this overriding URL is set. Unless I'm missing something?

gromdimon: Good notice! I've implemented that, but also found unwanted behaviour (which can not handle in a simple way):
If: 
- user provides `cleanup_method`, that does not change `url` + 
- there is reference `self.url` in that test + 
- user sets url param in `assert_response*` to some variable different from `self.url` 
- == we get bug :)

However, there's no logic in such a "developer experience", but any vulnerabilities are not welcome ((

mikkonie (changes requested):
**Edit:** I was a bit confused earlier, but trying this myself I see the issue: when iterating within the method, we still rely on the single URL provided. Somehow I became blind to the fact that we provide the URL to the method once and it then remains static regardless of what we update in `cleanup_method`. Ok, I finally get it :)

However, I still maintain this update should not be needed at all. Just using the cleanup method (which we'll need anyway) and setting a permanent UUID for the object works fine and we don't have to make the assertion method more complicated. [See e.g. here](https://github.com/bihealth/sodar-server/blob/dev/samplesheets/tests/test_permissions_api.py#L578). Or is there a special case where this somehow doesn't work and if so, what is it?

gromdimon:
You're right! The problem came from sodar-server, but since it was solved this issue is not valid anymore. Closing PR though

PR #1242, opened 2023-07-18
fix: using forks from pypi rather than source URLs in dependencies

holtgrewe:
We probably need this in the `main` branch as well.

https://pypi.org/project/django-plugins-bihealth/

https://pypi.org/project/drf-keyed-list-bihealth/

We'd appreciate a release of django-sodar-core `v0.11.2`


mikkonie:
Already done in the [0.11.2](https://github.com/bihealth/sodar-core/tree/0.11.2) branch and the corresponding release, closing.

PR #1279, opened 2023-09-11
Add common jquery and css class for submit button disabling

gromdimon:
### Tasks:
- [x] Implement
- [x] Test

### DIFF
--- a/projectroles/static/projectroles/js/projectroles.js
+++ b/projectroles/static/projectroles/js/projectroles.js
@@ -319,7 +319,7 @@ $(window).on('resize', function () {
 });
 
 
-/* Disable button once they clicked ----------------------------------------- */
+/* Disable submit button once clicked --------------------------------------- */
 
 
 $(document).ready(function () {
--- a/projectroles/static/projectroles/js/projectroles.js
+++ b/projectroles/static/projectroles/js/projectroles.js
@@ -317,3 +317,18 @@ $(document).ready(function () {
 $(window).on('resize', function () {
     toggleSidebar();
 });
+
+
+/* Disable button once they clicked ----------------------------------------- */
+
+
+$(document).ready(function () {
+    $('.sodar-pr-btn-submit-once').click(function () {
+        $(this).attr('disabled', true);
+        $(this).find('svg').remove();
+        $(this).prepend('<span class="spinner-border spinner-border-sm" ' +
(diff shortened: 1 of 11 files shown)


projectroles/static/projectroles/js/projectroles.js
```diff
@@ -317,3 +317,18 @@ $(document).ready(function () {
 $(window).on('resize', function () {
     toggleSidebar();
 });
+
+
+/* Disable button once they clicked ----------------------------------------- */
```
mikkonie: Maybe this should read "Disable submit button once clicked".

projectroles/tests/test_ui.py
```diff
@@ -1688,6 +1688,24 @@ def test_settings_label_icon(self):
         ).find_element(By.TAG_NAME, 'svg')
         self.assertTrue(logo.is_displayed())
 
+    def test_submit_button(self):
```
mikkonie, line 1696: This can be tested by looping the check for `is_enabled()` after clicking, perhaps with a short delay in between each try if needed. The loop should repeat until we either catch the expected result or time out after a maximum amount of retries.

If that doesn't work, let me know. There are other possibilities..

mikkonie (changes requested):
Good work, this looks and works really nicely! I had one comment and a suggestion on how to implement the test.

projectroles/tests/test_ui.py
```diff
@@ -1688,6 +1693,37 @@ def test_settings_label_icon(self):
...
+
+        self.assertTrue(element.is_enabled())
+
+        # Define timeout and retry interval
+        timeout = 10  # Maximum time to wait in seconds
+        retry_interval = 0.5  # Time between retries in seconds
+
+        start_time = time.time()
+        element.click()
+        while True:
```
mikkonie: This is looking good, I have some ideas for further improvement:

- Instead of calculating time, we can just set a number of retries, iterate with `range()` and `break` once condition is met. This would be more compact while doing the same thing.
- 0.5 seconds might be a bit too much to wait. I feel this operation might execute faster than that in some cases.. Maybe 0.2s would be better?
- As final cleanup, make sure to remove unneeded blank lines :)

mikkonie (changes requested):

mikkonie (approved):
Looking good, approved :)

Issue #1376, opened 2024-03-04
Update python installation utility script to support `python3.11`


gromdimon:
### Problem

After https://github.com/bihealth/sodar-core/commit/be012e5536bacf8bfbfe95e3c930324edae0309b implementation sodar-core environment uses python3.11. However the installation script still refer to python3.8.

### Solution

Adapt utility sctipts to meet new environment requirements.

### Alternative Solutions
N/A

### Additional Context
N/A

mikkonie:
Done by @gromdimon, merged.

Issue #1374, opened 2024-02-29
Upgrade depdendencies for v1.0.0


mikkonie:
To be done after merging #880.

- [x] General upgrade to all Python deps, unless serious breaking changes are detected
- [x] Latest flake8 & relinting
- [x] Latest black & reformatting
- [x] Fix selenium init changes
- [x] ~~Try out `django-iconify==0.4`~~ 0.4 requires Python>=3.10, but 0.3 works
- [x] Try out `sphinx-rtd-theme==2.0.0` and related dependencies
- [x] Try out latest `django-autocomplete-light` (see #1224)
- [x] ~~Try out upgrading `coveralls` and `coverage`~~ No updates for `coveralls`
- [x] Upgrade & reinstall versioneer
- [x] `Django>=4.2.14, <5.0`
- [x] `black>=24.3.0`
- [x] `djangorestframework>=3.15.2`
- [x] `setuptools>=70.0.0`

(Note: this will be kept open until the release of v1.0.0 to add new upgrades if needed)

mikkonie:
Done.

PR #1377, opened 2024-03-04
Update python installation script in order to accept python3.11 

gromdimon:
Just small changes for #1376

### DIFF
--- a/utility/install_os_gitlab.sh
+++ b/utility/install_os_gitlab.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Install OS dependencies for Docker image python:3.8 used in Gitlab CI
+# Install OS dependencies for Docker image python used in Gitlab CI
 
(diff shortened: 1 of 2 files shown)


gromdimon:
By the way, I found more artifacts, but I'm not sure if they need to be updated:
 - [ ] `./.readthedocs.yaml#L11`
- [ ] `..versioneer.py` - Just a general update of versioneer

mikkonie:
> By the way, I found more artifacts, but I'm not sure if they need to be updated:
> 
>     * [ ]  `./.readthedocs.yaml#L11`
> 
>     * [ ]  `..versioneer.py` - Just a general update of versioneer

Versioneer will be updated as a part of #1374, I'm doing that right now. Readthedocs build is not that critical, but I can also upgrade it as part of that issue.

mikkonie (approved):
Looks good, will merge :)

Issue #1378, opened 2024-03-04
Update postgres installation script to support `postgresql16`


gromdimon:
### Problem

After https://github.com/bihealth/sodar-core/commit/be012e5536bacf8bfbfe95e3c930324edae0309b implementation sodar-core environment uses postgresql-16. However the installation script still refer to postgresql-11

### Solution

Adapt utility sctipts to meet new environment requirements.

### Alternative Solutions
N/A

### Additional Context
N/A

mikkonie:
Done by @gromdimon, merged.

PR #1379, opened 2024-03-04
Update PostgreSQL installation script to support postgresql-16

gromdimon:
Just small changes for #1378


mikkonie (approved):
Looking good here as well.

Issue #1380, opened 2024-03-04
Refactor sidebar logic into reusable component


mikkonie:
In #1366, we have a use case for providing sidebar content via an Ajax view to e.g. client-side single page apps. This enables them to display the sidebar content as they choose, without requiring the use of base SODAR Core templates.

Right now, the logic for building the sidebar is completely contained in the `_project_sidebar.html` template and associated template tags. We should move the logic away from the template and provide the necessary data as a reusable helper component. That enables us to both render the list in the server-side template in SODAR Core, as well as provide it to client-side SPAs via the Ajax view.

Where to place the new helper turned out to be a somewhat complex issue. Having this as a context processor would not be optimal, as retrieving e.g. the project context there would be tricky. Also, this would create a very spagetti-ish processor method. My next idea was to extend `ProjectContextMixin`, but this would not correctly build the sidebar for site app views or the home view. The only 100% common factor between these views is the sidebar template, so we should still retrieve the list data via a template tag. But of course a template tag should not be used to add data into an Ajax view.

In the end, the following solution seems to make the most sense:

1. Add new utility class to build sidebar links
2. Move logic from template and template tags into utility class
3. Wrap utility class back into a single template tag
4. Render list in template based on data returned from utility class

## Spec

- Add `AppLinkContent` class to `projectroles.utils`
    * Why utils? Because it's not a mixin as we also have to access it directly in template tags.
    * Why a class? Because otherwise we're setting ourselves up for circular dependency hell.
    * This also will be used for the user dropdown links (#1381)
- Add method `get_sidebar_links()`
    * Args should be:
        + `request` (`HTTPRequest` object)
        + `project=None` (`Project` object or `None`)
    * This should return a list of dicts with each dict containing the following:
        + `url`: URL for the link
        + `label`: Text label to be displayed
        + `icon`: Icon to be used in format of `namespace:id`, e.g. `mdi:home`
        + `active`: Boolean for whether the link should be currently highlighted as active
- Move sidebar logic from template and template tags to new class
    * Project context checks, perm checks, visibility checks, state/activity checks, etc
    * Goal: all logic happens within `get_sidebar_links()` (and potential internal helper methods in the new class)
    * The method simply returns a list of links valid for the current view and user
- Add wrapper template tag in `projectroles_tags` to retrieve sidebar list
    * Call the template tag to retrieve the links into the template
        + Pass `project` from view context to parameter (will either be an object or None)
    * Still use values from `sidebar_processor.py`
[...]

PR #1384, opened 2024-03-06
Refactor sidebar logic into reusable component

gromdimon:
Issue: #1380

### DIFF
--- a/projectroles/templates/projectroles/_project_sidebar.html
+++ b/projectroles/templates/projectroles/_project_sidebar.html
@@ -3,18 +3,18 @@
 {# Project nav #}
 {% get_sidebar_links request project as sidebar_links %}
 {% for link in sidebar_links %}
-      <li
-          id="sodar-pr-nav-{{ link.name }}"
-          class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}">
-          <a class="nav-link"
-              href="{{ link.url }}"
-              id="sodar-pr-link-{{ link.name }}">
-              <span class="sodar-pr-sidebar-icon">
-                  <i class="iconify"
-                      data-icon="{{ link.icon }}"
-                      data-height="{{ sidebar_icon_size }}"></i>
-              </span>
-              <br />{{ link.label | safe }}
-          </a>
-      </li>
+  <li
+    id="sodar-pr-nav-{{ link.name }}"
+    class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}">
+    <a class="nav-link"
+      href="{{ link.url }}"
+      id="sodar-pr-link-{{ link.name }}">
+      <span class="sodar-pr-sidebar-icon">
+        <i class="iconify"
+            data-icon="{{ link.icon }}"
+            data-height="{{ sidebar_icon_size }}"></i>
+      </span>
+      <br />{{ link.label | safe }}
+    </a>
+  </li>
 {% endfor %}
--- a/projectroles/templates/projectroles/_project_sidebar.html
+++ b/projectroles/templates/projectroles/_project_sidebar.html
@@ -4,7 +4,8 @@
 {% get_project_app_links request project as sidebar_links %}
 {% for link in sidebar_links %}
   <li id="sodar-pr-nav-{{ link.name }}"
-      class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}">
+      class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}"
+      style="word-break: break-word;">
     <a class="nav-link"
       href="{{ link.url }}"
       id="sodar-pr-link-{{ link.name }}">
--- a/projectroles/templates/projectroles/_project_sidebar.html
+++ b/projectroles/templates/projectroles/_project_sidebar.html
@@ -1,18 +1,17 @@
 {% load projectroles_tags %}
 
 {# Project nav #}
-{% get_sidebar_links request project as sidebar_links %}
+{% get_project_app_links request project as sidebar_links %}
 {% for link in sidebar_links %}
-  <li
-    id="sodar-pr-nav-{{ link.name }}"
-    class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}">
+  <li id="sodar-pr-nav-{{ link.name }}"
+      class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}">
     <a class="nav-link"
       href="{{ link.url }}"
       id="sodar-pr-link-{{ link.name }}">
       <span class="sodar-pr-sidebar-icon">
         <i class="iconify"
-            data-icon="{{ link.icon }}"
-            data-height="{{ sidebar_icon_size }}"></i>
+           data-icon="{{ link.icon }}"
+           data-height="{{ sidebar_icon_size }}"></i>
       </span>
       <br />{{ link.label | safe }}
     </a>
--- a/projectroles/templates/projectroles/_project_sidebar.html
+++ b/projectroles/templates/projectroles/_project_sidebar.html
@@ -4,11 +4,11 @@
 {% get_sidebar_links request project as sidebar_links %}
 {% for link in sidebar_links %}
       <li
-          id="sodar-pr-nav-app-plugin-{{ link.name }}"
+          id="sodar-pr-nav-{{ link.name }}"
           class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}">
           <a class="nav-link"
               href="{{ link.url }}"
-              id="sodar-pr-link-app-plugin-{{ link.name }}">
+              id="sodar-pr-link-{{ link.name }}">
               <span class="sodar-pr-sidebar-icon">
                   <i class="iconify"
                       data-icon="{{ link.icon }}"
--- a/projectroles/templates/projectroles/_project_sidebar.html
+++ b/projectroles/templates/projectroles/_project_sidebar.html
@@ -1,146 +1,20 @@
-{% load rules %}
 {% load projectroles_tags %}
-{% load projectroles_common_tags %}
-
-{% sodar_constant 'PROJECT_TYPE_PROJECT' as PROJECT_TYPE_PROJECT %}
-{% sodar_constant 'PROJECT_TYPE_CATEGORY' as PROJECT_TYPE_CATEGORY %}
-{% allow_project_creation as allow_creation %}
-{% get_django_setting 'PROJECTROLES_DISABLE_CATEGORIES' as disable_categories %}
-{% get_display_name 'PROJECT' title=True as project_display %}
-{% get_display_name 'CATEGORY' title=True as category_display %}
 
 {# Project nav #}
-
-{% if project %}
-  {% has_perm 'projectroles.view_project' request.user project as can_view_project %}
-  {% has_perm 'projectroles.view_project_roles' request.user project as can_view_roles %}
-  {% has_perm 'projectroles.update_project' request.user project as can_update_project %}
-
-  {# Overview #}
-
-  <li id="sodar-pr-nav-project-detail"
-      class="nav-item sodar-pr-sidebar-nav-item {% get_pr_link_state projectroles_urls request 'detail' %}">
-    <a class="nav-link sodar-pr-sidebar-nav-link"
-       href="{% url 'projectroles:detail' project=project.sodar_uuid %}"
-       id="sodar-pr-link-project-detail">
-      <span class="sodar-pr-sidebar-icon">
-        {% if project.type == PROJECT_TYPE_CATEGORY %}
-          <i class="iconify"
-             data-icon="mdi:rhombus-split"
-             data-height="{{ sidebar_icon_size }}"></i>
-        {% else %}
-          <i class="iconify"
-             data-icon="mdi:cube"
-             data-height="{{ sidebar_icon_size }}"></i>
-        {% endif %}
-      </span>
-      <br />{% get_display_name project.type title=True %}<br />Overview
-    </a>
-  </li>
-
-  {# App plugins #}
-  {% for plugin in app_plugins %}
-    {% is_app_visible plugin project request.user as app_link_visible %}
-    {% if app_link_visible %}
-      {% get_sidebar_app_legend plugin.title as app_legend %}
-      <li id="sodar-pr-nav-app-plugin-{{ plugin.name }}"
-          class="nav-item sodar-pr-sidebar-nav-item {% get_app_link_state plugin request.resolver_match.app_name request.resolver_match.url_name %}">
-        <a class="nav-link"
-           href="{% url plugin.entry_point_url_id project=project.sodar_uuid %}"
-           id="sodar-pr-link-app-plugin-{{ plugin.name }}">
-         <span class="sodar-pr-sidebar-icon">
-           <i class="iconify"
-              data-icon="{{ plugin.icon }}"
-              data-height="{{ sidebar_icon_size }}"></i>
-         </span>
-         <br />{{ app_legend | safe }}
-        </a>
+{% get_sidebar_links request project as sidebar_links %}
+{% for link in sidebar_links %}
+      <li
+          id="sodar-pr-nav-app-plugin-{{ link.name }}"
+          class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}">
+          <a class="nav-link"
+              href="{{ link.url }}"
+              id="sodar-pr-link-app-plugin-{{ link.name }}">
+              <span class="sodar-pr-sidebar-icon">
+                  <i class="iconify"
+                      data-icon="{{ link.icon }}"
+                      data-height="{{ sidebar_icon_size }}"></i>
+              </span>
+              <br />{{ link.label | safe }}
+          </a>
       </li>
-    {% endif %}
-  {% endfor %}
-
-  {# Role and project editing #}
-  {% if can_view_roles %}
-    <li id="sodar-pr-nav-project-roles"
-        class="nav-item sodar-pr-sidebar-nav-item {% get_pr_link_state projectroles_urls request role_urls %}">
-      <a class="nav-link"
-         href="{% url 'projectroles:roles' project=project.sodar_uuid %}"
-         id="sodar-pr-link-project-roles">
-        <span class="sodar-pr-sidebar-icon">
-          <i class="iconify"
-             data-icon="mdi:account-multiple"
-             data-height="{{ sidebar_icon_size }}"></i>
-          <br />Members
-        </span>
-      </a>
-    </li>
-  {% endif %}
-  {% if can_update_project %}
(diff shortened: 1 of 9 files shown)


gromdimon:
It seems, that there's no template tags which can be removed after refactoring. However, I found that `projectroles_tags.get_role_import_action` is not used anywhere. Is it some kind of an artifact or or it's used somewhere in sodar-server/varfish or I missed something?

btw. this template_tag was not used in `_project_sidebar.html`

mikkonie:
> It seems, that there's no template tags which can be removed after refactoring. However, I found that `projectroles_tags.get_role_import_action` is not used anywhere. Is it some kind of an artifact or or it's used somewhere in sodar-server/varfish or I missed something?

Nice find. This seems to be an old template tag related to remote project sync which was refactored away, but I guess we forgot to remove the tag. It shouldn't be used anywhere else either, as it's in the `projectroles_tags` module, not the `projectroles_common_tags` one. Feel free to remove. It can be done as part of this branch even if not directly related.

projectroles/templates/projectroles/_project_sidebar.html
```diff
@@ -1,146 +1,20 @@
...
-         <span class="sodar-pr-sidebar-icon">
-           <i class="iconify"
-              data-icon="{{ plugin.icon }}"
-              data-height="{{ sidebar_icon_size }}"></i>
-         </span>
-         <br />{{ app_legend | safe }}
-        </a>
+{% get_sidebar_links request project as sidebar_links %}
+{% for link in sidebar_links %}
+      <li
```
mikkonie: Minor style notes for templates: 1) make sure to use 2-space indent and 2) have the first attribute(s) on the same line as the tag opening.

(And yes, we should probably look into integrating djlint or something similar. Last time I looked there weren't really many decent Django template formatters around, but then again that was 5+ years ago..)

projectroles/utils.py
```diff
@@ -95,3 +109,197 @@ def get_app_names():
             else:
                 ret.append(s[0])
     return sorted(ret)
+
+
+class SidebarContent:
+    """Class for generating sidebar content for the site"""
+
+    def _is_active(self, request, link_names=None):
```
mikkonie: The naming of this method is confusing, as it specifically checks the state of links for the projectroles app. Perhaps `_is_active_projectroles()` would be more descriptive.

Also, we are now duplicating a lot the code in the `get_pr_link_state()` template tag. That is also used in the `_project_menu_btn.html` template. We need to refactor that template to also use this helper class. I will add a separate comment about this and update the spec.

projectroles/utils.py
```diff
@@ -95,3 +109,197 @@ def get_app_names():
...
+        Check if current URL is active in the sidebar for a specific app plugin.
+        """
+        if app_plugin.name.startswith(app_name) and url_name in [
+            u.name for u in getattr(app_plugin, 'urls', [])
+        ]:
+            return True
+        # HACK for remote site views, see issue #1336
+        if (
+            app_name == 'projectroles'
+            and app_plugin.name == 'remotesites'
```
mikkonie, line 141: Note that the `get_app_link_state()` template tag is used for both sidebar and user dropdown links. The remote sites part here only applies to the user dropdown.

We shouldn't have needless repetition either. The optimal thing would be to have a single helper class for both the sidebar and the user dropdown retrieval, where both API methods would use this `_is_active_plugin()` helper. This is something I overlooked in my original specs..

projectroles/utils.py
```diff
@@ -95,3 +109,197 @@ def get_app_names():
...
+        """Check if current URL is active in the sidebar."""
+        # HACK: Avoid circular import
+        from projectroles.urls import urlpatterns
+
+        if request.resolver_match.app_name != 'projectroles':
+            return False
+        url_name = request.resolver_match.url_name
+        if url_name in [u.name for u in urlpatterns]:
+            if link_names:
+                if not isinstance(link_names, list):
```
mikkonie: Since we now always call this method in python instead of templates, we can get rid of this hack. Just provide the `link_names` arg as a list in each call and we don't need to do a conversion.

projectroles/utils.py
```diff
@@ -95,3 +109,197 @@ def get_app_names():
...
+        app_hidden = False
+        if plugin.name in getattr(
+            settings, 'PROJECTROLES_HIDE_PROJECT_APPS', []
+        ):
+            app_hidden = True
+        if (
+            can_view_app
+            and not app_hidden
+            and (
+                project.type == SODAR_CONSTANTS['PROJECT_TYPE_PROJECT']
```
mikkonie: Hint: you can define "shortcuts" like `PROJECT_TYPE_PROJECT = SODAR_CONSTANTS['PROJECT_TYPE_PROJECT']` in the module root to shorten constant checks like this nicely.

projectroles/utils.py
```diff
@@ -95,3 +109,197 @@ def get_app_names():
...
+        if project:
+            # Add project overview link
+            links.append(
+                {
+                    'name': 'project-detail',
+                    'url': reverse(
+                        'projectroles:detail',
+                        kwargs={'project': project.sodar_uuid},
+                    ),
+                    'label': 'Project<br />Overview',
```
mikkonie: As I mentioned before we shouldn't use these `<br />` tags, at leat not here where we simply return data. These need to be line-broken by CSS in rendering.

projectroles/utils.py
```diff
@@ -95,3 +109,197 @@ def get_app_names():
...
+            settings.PROJECTROLES_SITE_MODE
+            == SODAR_CONSTANTS['SITE_MODE_TARGET']
+            and not settings.PROJECTROLES_TARGET_CREATE
+        ):
+            return False
+        return True
+
+    def get_sidebar_links(self, request, project=None):
+        """Return sidebar links based on the current project and user."""
+        links = []
```
mikkonie: Minor note, the conventional shorthand for return data variables is `ret`. (Don't ask me where that came from, can't remember!)

projectroles/tests/test_templatetags.py
```diff
@@ -618,3 +618,75 @@ def test_get_sidebar_app_legend(self):
         self.assertEqual(
             tags.get_sidebar_app_legend('Update Project'), 'Update<br />Project'
         )
+
+    def test_get_sidebar_links_homepage(self):
```
mikkonie: "Homepage" isn't a standard term in the project. "Home view" can be used.

We should also have tests for other resolvers, such as the project detail view, role views (to ensure the active checks work) and some app view (example project app is usually used for this).

projectroles/tests/test_templatetags.py
```diff
@@ -618,3 +618,75 @@ def test_get_sidebar_app_legend(self):
...
+        """Test get_sidebar_links() on the homepage"""
+        url = reverse('home')
+        req_factory = RequestFactory()
+        with self.login(self.user):
+            request = req_factory.get(url)
+            request.resolver_match = resolve(url)
+            request.user = self.user
+            self.assertEqual(
+                tags.get_sidebar_links(request, self.project),
```
mikkonie: This expected result is in fact wrong. In the home view, we get no sidebar links, except for superusers who get the "create category" link. I've tested the implementation and it works correctly, so this test must be malfunctioning. For one, there is no project context in the home view so we should not be providing `self.project` to the method.

projectroles/utils.py
```diff
@@ -95,3 +109,197 @@ def get_app_names():
             else:
                 ret.append(s[0])
     return sorted(ret)
+
+
+class SidebarContent:
```
mikkonie: I think we need to unify the sidebar and user dropdown helpers into a single class, see my comment further on. Perhaps the naming could be something like `AppLinkContent`.

mikkonie (changes requested):
The implementation is working correctly, good work. Here are some comments. There is one major thing I missed in my spec, I will write a separate comment about that and also update the spec.

mikkonie:
Ok, so: I completely overlooked `_project_menu_btn.html` in my spec. It is basically a responsive replacement for the sidebar, containing the exact same links but rendered as a dropdown.

Needless to say, this should also be refactored to use the new helper class instead of having the logic within its template. So here's what should be done:

1. Refactor `_project_menu_btn.html` to get links from the helper and render them as is done currently
2. Remove unused template tags

projectroles/utils.py
gromdimon: Yeah, totally forgot! Actually, how are we going to change ` ` to `<br />` in css? I still don't get it (

projectroles/templates/projectroles/_project_sidebar.html
```diff
@@ -1,146 +1,20 @@
...
-      </a>
-    </li>
-  {% endif %}
-{# Allow project creation under root #}
-{% elif disable_categories and request.user.is_superuser %}
-  <li id="sodar-pr-nav-project-create"
-      class="nav-item sodar-pr-sidebar-nav-item {% get_pr_link_state projectroles_urls request 'create' %}">
+{% get_sidebar_links request project as sidebar_links %}
+{% for link in sidebar_links %}
+  <li
```
mikkonie: Here we still have a line break after the tag opening. Let's have the first attribute on the same line instead to save up space and remain consistent.

projectroles/templatetags/projectroles_tags.py
```diff
@@ -285,3 +268,10 @@ def get_admin_warning():
         '</a></p>'.format(reverse('admin:index'))
     )
     return ret
+
+
+@register.simple_tag
+def get_sidebar_links(request, project=None):
```
mikkonie: Since this is used for both the sidebar and the dropdown, we can rename it to `get_project_app_links()`.

projectroles/templatetags/projectroles_tags.py
```diff
@@ -285,3 +268,10 @@ def get_admin_warning():
...
+    """Return sidebar links"""
+    sidebar_content = AppLinkContent()
```
mikkonie: Here we can also rename this as `app_links` or something.

projectroles/utils.py
```diff
@@ -15,6 +15,22 @@
 # SODAR constants
 SODAR_CONSTANTS = get_sodar_constants()
```
mikkonie: You should instead import `SODAR_CONSTANTS` from `projectroles.models`. See e.g. `projectroles.views` for an example.

projectroles/utils.py
```diff
@@ -15,6 +15,22 @@
...
 
+# Local constants
+PROJECT_TYPE_PROJECT = SODAR_CONSTANTS['PROJECT_TYPE_PROJECT']
```
mikkonie, line 16: These should go under `# SODAR constants`. Again, see `projectroles.views` for how this is usually done.

projectroles/utils.py
```diff
@@ -95,3 +111,195 @@ def get_app_names():
             else:
                 ret.append(s[0])
     return sorted(ret)
+
+
+class AppLinkContent:
+    """Class for generating sidebar content for the site"""
```
mikkonie: This docstring could be updated since we also use this for the project dropdown as well as (in the future) in the user dropdown. "Class for generating application links for the UI" would work.

projectroles/utils.py
```diff
@@ -95,3 +111,195 @@ def get_app_names():
...
+        """Check if current URL is active in the sidebar."""
+        # HACK: Avoid circular import
+        from projectroles.urls import urlpatterns
+
+        if request.resolver_match.app_name != 'projectroles':
+            return False
+        url_name = request.resolver_match.url_name
+        if url_name in [u.name for u in urlpatterns]:
+            if link_names:
+                if url_name not in link_names:
```
mikkonie: I believe these two nested `if`statements can be combined into one.

projectroles/utils.py
```diff
@@ -95,3 +111,195 @@ def get_app_names():
...
+                    return False
+            return True
+        return False
+
+    def _is_active_plugin(self, app_plugin, app_name, url_name):
+        """
+        Check if current URL is active in the sidebar for a specific app plugin.
```
mikkonie: The sidebar mention can be removed for this docstring, as this function is also used elsewhere.

projectroles/utils.py
```diff
@@ -95,3 +111,195 @@ def get_app_names():
             else:
                 ret.append(s[0])
     return sorted(ret)
+
+
+class AppLinkContent:
+    """Class for generating sidebar content for the site"""
+
+    def _is_active_projectroles(self, request, link_names=None):
+        """Check if current URL is active in the sidebar."""
```
mikkonie: A more relevant docstring would be "Check if current URL is active under the projectroles app".

projectroles/utils.py
```diff
@@ -95,3 +111,195 @@ def get_app_names():
...
+        if (
+            settings.PROJECTROLES_SITE_MODE
+            == SODAR_CONSTANTS['SITE_MODE_TARGET']
+            and not settings.PROJECTROLES_TARGET_CREATE
+        ):
+            return False
+        return True
+
+    def get_app_links(self, request, project=None):
+        """Return sidebar links based on the current project and user."""
```
mikkonie: Again we can just replace "sidebar links" with "project app links".

projectroles/templates/projectroles/_project_sidebar.html
```diff
@@ -1,146 +1,20 @@
...
-       href="{% url 'projectroles:create' %}"
-       id="sodar-pr-link-project-create">
+      href="{{ link.url }}"
+      id="sodar-pr-link-{{ link.name }}">
       <span class="sodar-pr-sidebar-icon">
         <i class="iconify"
-           data-icon="mdi:plus-thick"
-           data-height="{{ sidebar_icon_size }}"></i>
+            data-icon="{{ link.icon }}"
+            data-height="{{ sidebar_icon_size }}"></i>
```
mikkonie: These two lines seem to have an extra whitespace in front of them.

projectroles/tests/test_templatetags.py
```diff
@@ -618,3 +618,211 @@ def test_get_sidebar_app_legend(self):
         self.assertEqual(
             tags.get_sidebar_app_legend('Update Project'), 'Update<br />Project'
         )
+
+    def test_get_sidebar_links_home_view(self):
```
mikkonie: `_home` is probably enough here since the "view" part is self-explanatory.

projectroles/tests/test_templatetags.py
```diff
@@ -618,3 +618,211 @@ def test_get_sidebar_app_legend(self):
...
+        with self.login(self.user):
+            request = req_factory.get(url)
+            request.resolver_match = resolve(url)
+            request.user = self.user
+            self.assertEqual(
+                tags.get_sidebar_links(request),
+                [],
+            )
+
+    def test_get_sidebar_links_project_view(self):
```
mikkonie: This is specifically for the project detail view, so `_project_detail` would be the appropriate name. Same for the docstring. Also note that we want to make unit test docstrings as concise as possible, so omitting e.g. articles is fine.

projectroles/tests/test_templatetags.py
```diff
@@ -618,3 +618,211 @@ def test_get_sidebar_app_legend(self):
...
+        """Test get_sidebar_links() on the role view"""
+        url = reverse(
+            'projectroles:roles', kwargs={'project': self.project.sodar_uuid}
+        )
+        req_factory = RequestFactory()
+        with self.login(self.user):
+            request = req_factory.get(url)
+            request.resolver_match = resolve(url)
+            request.user = self.user
+            self.assertEqual(
```
mikkonie: It's enough if we test the `active` key of each link here.

projectroles/tests/test_templatetags.py
```diff
@@ -618,3 +618,211 @@ def test_get_sidebar_app_legend(self):
...
+        url = reverse(
+            'timeline:list_project', kwargs={'project': self.project.sodar_uuid}
+        )
+        req_factory = RequestFactory()
+        with self.login(self.user):
+            request = req_factory.get(url)
+            request.resolver_match = resolve(url)
+            request.user = self.user
+            self.assertEqual(
+                tags.get_sidebar_links(request, self.project),
```
mikkonie: Same here, we only need to test the `active` values here.

mikkonie (changes requested):
This is proceeding nicely, I had some more feedback.

projectroles/templates/projectroles/_project_sidebar.html
```diff
@@ -1,146 +1,20 @@
...
+      href="{{ link.url }}"
+      id="sodar-pr-link-{{ link.name }}">
       <span class="sodar-pr-sidebar-icon">
         <i class="iconify"
-           data-icon="mdi:plus-thick"
+           data-icon="{{ link.icon }}"
            data-height="{{ sidebar_icon_size }}"></i>
       </span>
-      <br />Create<br />{{ project_display }}
+      <br />{{ link.label | safe }}
```
mikkonie: As I mentioned in the previous review, the `safe` filter is not needed here as the label is not expected to contain any HTML.

projectroles/tests/test_templatetags.py
```diff
@@ -618,3 +618,113 @@ def test_get_sidebar_app_legend(self):
         self.assertEqual(
             tags.get_sidebar_app_legend('Update Project'), 'Update<br />Project'
         )
+
+    def test_get_sidebar_links_home(self):
+        """Test get_sidebar_links() on the home view"""
+        url = reverse('home')
+        req_factory = RequestFactory()
```
mikkonie: We can create this in `setUp()` as `self.req_factory` instead, so we'll save a few lines of code in re-initing it inside the test case methods.

projectroles/utils.py
```diff
@@ -95,3 +110,192 @@ def get_app_names():
...
+        """Check whether creating a project is allowed on the site."""
+        if (
+            settings.PROJECTROLES_SITE_MODE
+            == SODAR_CONSTANTS['SITE_MODE_TARGET']
+            and not settings.PROJECTROLES_TARGET_CREATE
+        ):
+            return False
+        return True
+
+    def get_app_links(self, request, project=None):
```
mikkonie: Let's call this `get_project_app_links()` too, as we'll be adding a second method for the site apps for the user dropdown in the other issue.

mikkonie (changes requested):
Good work! I had just a couple very minor comments, after these I'm happy to merge.

projectroles/templates/projectroles/_project_sidebar.html
```diff
@@ -1,146 +1,20 @@
...
-  {% endif %}
-{# Allow project creation under root #}
-{% elif disable_categories and request.user.is_superuser %}
-  <li id="sodar-pr-nav-project-create"
-      class="nav-item sodar-pr-sidebar-nav-item {% get_pr_link_state projectroles_urls request 'create' %}">
+{% get_project_app_links request project as sidebar_links %}
+{% for link in sidebar_links %}
+  <li id="sodar-pr-nav-{{ link.name }}"
+      class="nav-item sodar-pr-sidebar-nav-item {% if link.active %}active{% endif %}"
+      style="word-break: break-word;">
```
mikkonie: Add this in `projectroles.css` under `.sodar-pr-sidebar-nav-item a`. We should not use `style` tags except in some very special cases where they're absolutely required.

mikkonie (approved):

PR #1386, opened 2024-03-06
Refactor user dropdown logic into reusable component

### LINKED ISSUE #1381: Refactor user dropdown logic into reusable component
Sequel to #1380, also needed for #1366.

Similar to the sidebar, we want to be able to provide the user dropdown links also in an Ajax view for possible client side UIs. And similar to the sidebar, the logic for the dropdown is contained in the context processor, template and template tags.

So, you guessed it: we need to refactor the user dropdown in a similar way to the sidebar.

## Spec

- Add `get_user_links()` method to `AppLinkContent` helper created in #1380
    * `request` should probably be the only arg required
    * Return list of dicts similar to the method in #1380
- Move logic from `_site_titlebar_dropdown.html` and related template tags into new method
- Deprecate `site_app_processor()`
- Add wrapper template tag to retrieve list into `_site_titlebar_dropdown.html`
    * Call it and render the dropdown links in the template
- Add tests for new helper class
- Add/update UI tests
    * I'm not sure if we have UI tests for the menu? If not, a new class should be added next to `TestProjectSidebar`. Testing with both regular user and admin
- Remove unused template tags (if any)

## Tasks

- [ ] Add helper class
- [ ] Move user dropdown logic into class
- [ ] Deprecate `site_app_processor`
- [ ] Update tests
- [ ] Remove unused template tags (if any)

gromdimon:
Issue: #1381

### DIFF
--- a/projectroles/templates/projectroles/_site_titlebar_dropdown.html
+++ b/projectroles/templates/projectroles/_site_titlebar_dropdown.html
@@ -50,8 +50,8 @@
       <a class="dropdown-item {% if link.active %}active{% endif %}
                 {% if link.name == 'sign-out' %}text-danger{% endif %}
                 {% if link.name == 'sign-in' %}text-primary{% endif %}"
-        href="{{ link.url }}"
-        id="sodar-navbar-link-{{ link.name }}">
+         href="{{ link.url }}"
+         id="sodar-navbar-link-{{ link.name }}">
         <i class="iconify" data-icon="{{ link.icon }}"></i> {{ link.label }}
       </a>
     {% endfor %}
--- a/projectroles/templates/projectroles/_site_titlebar_dropdown.html
+++ b/projectroles/templates/projectroles/_site_titlebar_dropdown.html
@@ -44,6 +44,9 @@
 
     {% get_user_links request as dropdown_links %}
     {% for link in dropdown_links %}
+      {% if link.name == 'admin' or link.name == 'sign-out' or link.name == 'sign-in' %}
+        <div class="dropdown-divider"></div>
+      {% endif %}
       <a class="dropdown-item {% if link.active %}active{% endif %}
                 {% if link.name == 'sign-out' %}text-danger{% endif %}
                 {% if link.name == 'sign-in' %}text-primary{% endif %}"
--- a/projectroles/templates/projectroles/_site_titlebar_dropdown.html
+++ b/projectroles/templates/projectroles/_site_titlebar_dropdown.html
@@ -5,48 +5,6 @@
 {% get_django_setting 'PROJECTROLES_KIOSK_MODE' as kiosk_mode %}
 
 {# Responsive replacement for user dropdown #}
-
-{# Admin link #}
-{#{% if request.user.is_superuser %}#}
-{#  <li class="nav-item sodar-navbar-alt-item">#}
-{#    <a class="nav-link" href="{% url 'admin:index' %}"#}
-{#       id="sodar-navbar-link-admin" target="_blank">#}
-{#      <i class="iconify" data-icon="mdi:cogs"></i> Admin#}
-{#    </a>#}
-{#  </li>#}
-{#{% endif %}#}
-{##}
-{# Site-wide apps #}
-{#{% for plugin in site_apps %}#}
-{#  {% has_perm plugin.app_permission request.user as can_view_app %}#}
-{#  {% if not plugin.app_permission or can_view_app %}#}
-{#    <li class="nav-item sodar-navbar-alt-item {% get_app_link_state plugin request.resolver_match.app_name request.resolver_match.url_name %}">#}
-{#      <a class="nav-link"#}
-{#         href="{% url plugin.entry_point_url_id %}"#}
-{#         id="sodar-navbar-siteapp-link-{{ plugin.name }}">#}
-{#        <i class="iconify" data-icon="{{ plugin.icon }}"></i> {{ plugin.title }}</a>#}
-{#    </li>#}
-{#  {% endif %}#}
-{#{% endfor %}#}
-{##}
-{# Log out link #}
-{#{% if request.user.is_authenticated %}#}
-{#  <li class="nav-item sodar-navbar-alt-item">#}
-{#    <a class="nav-link text-danger"#}
-{#       href="{% url 'logout' %}"#}
-{#       id="sodar-alt-nav-link-sign-out">#}
-{#      <i class="iconify" data-icon="mdi:logout-variant"></i> Logout#}
-{#    </a>#}
-{#  </li>#}
-{#{% elif not kiosk_mode %}#}
-{#  <li class="nav-item sodar-navbar-alt-item">#}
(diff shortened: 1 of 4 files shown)


mikkonie (approved):

projectroles/templates/projectroles/_site_titlebar_dropdown.html
```diff
@@ -75,45 +42,18 @@
...
-    {% if request.user.is_authenticated %}
-      <a class="dropdown-item text-danger"
-         href="{% url 'logout' %}"
-         id="sodar-navbar-link-logout">
-        <i class="iconify" data-icon="mdi:logout-variant"></i> Logout
+      <a class="dropdown-item {% if link.active %}active{% endif %}
+                {% if link.name == 'sign-out' %}text-danger{% endif %}
+                {% if link.name == 'sign-in' %}text-primary{% endif %}"
+        href="{{ link.url }}"
+        id="sodar-navbar-link-{{ link.name }}">
```
mikkonie: Super minor note, but these lines are missing one extra space in front of them :)

projectroles/templatetags/projectroles_tags.py
```diff
@@ -8,7 +8,7 @@
 from projectroles.app_settings import AppSettingAPI
 from projectroles.models import RemoteProject, SODAR_CONSTANTS
 from projectroles.plugins import get_active_plugins
-from projectroles.utils import AppLinkContent
+from projectroles.utils import UserDropdownContent, AppLinkContent
```
mikkonie: Like I specified earlier, the helpers for the user dropdown should go into `AppLinkContent` instead of their own class.

projectroles/utils.py
```diff
@@ -299,3 +299,60 @@ def get_project_app_links(self, request, project=None):
                 }
             )
         return ret
+
+
+class UserDropdownContent:
+    """Class for generating application links for the dropdown"""
+
+    def get_user_links(self, request):
```
mikkonie, line 303: Move this under `AppLinkContent` so we don't need a second class.

projectroles/utils.py
```diff
@@ -299,3 +299,60 @@ def get_project_app_links(self, request, project=None):
...
+                app.app_permission
+            ):
+                links.append(
+                    {
+                        'name': app.name,
+                        'url': reverse(app.entry_point_url_id),
+                        'label': app.title,
+                        'icon': app.icon,
+                        'active': request.path
+                        == reverse(app.entry_point_url_id),
```
mikkonie: For this, we need to use the `_is_active_plugin()` helper in `AppLinkContent`. See the original implementation. That's where we do the hack for remote site URLs under the projectroles app. Currently in this implementation, the UI does not highlight remote site views beyond the plugin entry point. See the following screenshot: `Remote Site Access` should be highlighed:

2024-03_remote_highlight_bug

projectroles/utils.py
```diff
@@ -299,3 +299,60 @@ def get_project_app_links(self, request, project=None):
...
             )
         return ret
+
+
+class UserDropdownContent:
+    """Class for generating application links for the dropdown"""
+
+    def get_user_links(self, request):
+        """Return user links for the user dropdown"""
+        links = []
```
mikkonie: Also here we should use the standard `ret` variable name.

projectroles/tests/test_templatetags.py
```diff
@@ -725,3 +725,71 @@ def test_get_sidebar_links_timeline_view(self):
                     self.assertEqual(app['active'], True)
                 else:
                     self.assertEqual(app['active'], False)
+
+    def test_get_get_user_links_home(self):
```
mikkonie: `get` seems to be repeated redundantly in the test name :)

projectroles/tests/test_templatetags.py
```diff
@@ -725,3 +725,71 @@ def test_get_sidebar_links_timeline_view(self):
...
+                        'name': 'sign-out',
+                        'url': '/logout/',
+                        'label': 'Logout',
+                        'icon': 'mdi:logout-variant',
+                        'active': False,
+                    },
+                ],
+            )
+
+    def test_get_user_links_userprofile(self):
```
mikkonie, line 878: Please also add further tests:

- Test home view as superuser, to assert all the superuser-only apps are visible
- Add test for `active` with a remote site url other than `projectroles:remote_sites` to ensure `_is_active_plugin()` is used correctly. You can use e.g. `projectroles:remote_site_create`.

mikkonie (changes requested):
Good work. I found one bug and have some requests for refactoring and minor cleanup.

projectroles/templatetags/projectroles_tags.py
```diff
@@ -270,6 +270,13 @@ def get_admin_warning():
     return ret
 
 
+@register.simple_tag
+def get_user_links(request):
+    """Return user dropdown links"""
+    app_links = AppLinkContent()
```
mikkonie: One minor optimization we could do: move this initialization from this method and `get_project_app_links()` into the root of the module (after `register = template.Library()`). This saves us from having to init the class multiple times. Probably a theoretical performance increase at best, but at least it will remove one redundant line of code :)

projectroles/tests/test_templatetags.py
```diff
@@ -725,3 +725,184 @@ def test_get_sidebar_links_timeline_view(self):
                     self.assertEqual(app['active'], True)
                 else:
                     self.assertEqual(app['active'], False)
+
+    def test_get_user_links_home(self):
```
mikkonie: Also one very small thing: for consistency, let's name these test methods in a kind of a hierarchical fashion. `get_links_home_user()` and `get_links_home_superuser()` etc is better, as the modifying things come last.

projectroles/tests/test_templatetags.py
```diff
@@ -725,3 +725,184 @@ def test_get_sidebar_links_timeline_view(self):
...
+        """Test get_user_links() on the home view"""
+        url = reverse('home')
+        with self.login(self.user):
```
mikkonie: With `RequestFactory`, using `self.login()` does nothing, as we create the request manually. Hence this can be removed from all tests.

projectroles/tests/test_templatetags.py
```diff
@@ -725,3 +725,184 @@ def test_get_sidebar_links_timeline_view(self):
...
+                        'icon': 'mdi:logout-variant',
+                        'active': False,
+                    },
+                ],
+            )
+
+    def test_get_superuser_links_home(self):
+        """Test get_user_links() on the home view as superuser"""
+        url = reverse('home')
+        superuser = self.make_user('superuser')
```
mikkonie: We can create the superuser in `setUp()`, to save us from several lines of re-init code in the test cases.

mikkonie (changes requested):
Good work again, had just a few small comments. After fixing these I'm sure we can merge :)

mikkonie (approved):
Everything looking good, this can be merged.

PR #1420, opened 2024-04-25
Add ajax views for retrieving sidebar and user dropdown links

### LINKED ISSUE #1366: Add Ajax views for retrieving sidebar and user dropdown links
In Varfish there is a case where several new views are implemented as single page apps and they don't really work with the base SODAR Core template.

It was requested to add Ajax views to provide the content for the sidebar and the user dropdown, so they can be rendered as suits best by the SPA. This can also be useful in other cases in the future where we may have views or entire sites which forego the base server side template.

Naturally, we want to reuse existing server side context code and wrap it in Ajax views.

We'll need to return the following:

- Link icon (as `namespace:id`, e.g. `mdi:home`)
- Text legend
- Link URL
- Boolean for "active"

## Spec

- Do #1380 and #1381
- Add Ajax view `SidebarContentAjaxView` in projectroles
    * URL name = `ajax_sidebar`
    * `project:uuid` given in the URL (no reason to call this without a project context)
    * Implement logic in `get()`
        + Use the `AppLinkContent' helper created in #1380 and #1381
        + Provide project to helper if found in URL kwargs
        + Also call `sidebar_processor()` and add its results into the return data
    * Challenge: where to get the current server-side URL to determine the `active` state?
        + The request URL will, of course, be the URL to the Ajax view itself. So that doesn't help..
        + In my opinion we have two options:
            + Figure out how to get the current URL the browser is on from the client (either from the request somehow or by an explicit parameter provided by the client to the Ajax view call)
            + Or, skip `active` for the Ajax views altogether: add a parameter to leave it out from the `AppLinkContent` methods
        + Let me know when you reach this point and if you'll figure out a nice way to implement it
- Add Ajax view `UserDropdownContentAjaxView` in projectroles
    * URL name = `ajax_user_dropdown`
    * No project context (irrelevant for user dropdown as we only show site/admin apps)
    * Use the `AppLinkContent` helper again
- Add documentation via docstrings and sphinx
- Add tests

## Tasks

- [x] Add `SidebarContentAjaxView`
- [x] Add `UserDropdownContentAjaxView`
- [x] Document
- [x] Test

gromdimon:
Issue: #1366

### DIFF
--- a/projectroles/tests/test_permissions_ajax.py
+++ b/projectroles/tests/test_permissions_ajax.py
@@ -263,7 +263,7 @@ class TestSidebarContentAjaxView(ProjectPermissionTestBase):
             self.user_contributor,
             self.user_guest,
         ]
-        bad_users = [self.user_no_roles, self.user_finder_cat, self.anonymous]
+        bad_users = [self.user_finder_cat, self.user_no_roles, self.anonymous]
         self.assert_response(self.url, good_users, 200)
         self.assert_response(self.url, bad_users, 403)
         self.project.set_public()
@@ -289,7 +289,7 @@ class TestSidebarContentAjaxView(ProjectPermissionTestBase):
             self.user_contributor,
             self.user_guest,
         ]
-        bad_users = [self.user_no_roles, self.user_finder_cat, self.anonymous]
+        bad_users = [self.user_finder_cat, self.user_no_roles, self.anonymous]
         self.assert_response(self.url, good_users, 200)
         self.assert_response(self.url, bad_users, 403)
 
--- a/projectroles/tests/test_permissions_ajax.py
+++ b/projectroles/tests/test_permissions_ajax.py
@@ -264,16 +264,34 @@ class TestSidebarContentAjaxView(ProjectPermissionTestBase):
             self.user_guest,
         ]
         bad_users = [self.user_no_roles, self.user_finder_cat, self.anonymous]
-        self.assert_response(self.url, good_users, 200, method='GET')
-        self.assert_response(self.url, bad_users, 403, method='GET')
+        self.assert_response(self.url, good_users, 200)
+        self.assert_response(self.url, bad_users, 403)
         self.project.set_public()
-        self.assert_response(self.url, self.user_no_roles, 200, method='GET')
+        self.assert_response(self.url, self.user_no_roles, 200)
 
     @override_settings(PROJECTROLES_ALLOW_ANONYMOUS=True)
     def test_get_anon(self):
         """Test GET with anonymous access"""
         self.project.set_public()
-        self.assert_response(self.url, self.anonymous, 200, method='GET')
+        self.assert_response(self.url, self.anonymous, 200)
+
+    def test_get_archive(self):
+        """Test SidebarContentAjaxView GET with archived project"""
+        self.project.set_archive()
+        good_users = [
+            self.superuser,
+            self.user_owner_cat,
+            self.user_delegate_cat,
+            self.user_contributor_cat,
+            self.user_guest_cat,
+            self.user_owner,
+            self.user_delegate,
+            self.user_contributor,
+            self.user_guest,
+        ]
+        bad_users = [self.user_no_roles, self.user_finder_cat, self.anonymous]
+        self.assert_response(self.url, good_users, 200)
+        self.assert_response(self.url, bad_users, 403)
 
     def test_get_category(self):
         """Test GET with category"""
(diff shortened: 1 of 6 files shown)


gromdimon:
- [x] Added ajax views
- [x] Added urls
- [x] Added tests
- [x] Added docstrings for usage

projectroles/views_ajax.py
```diff
@@ -431,6 +431,51 @@ def post(self, request, *args, **kwargs):
...
+    as GET parameters. Both of them refer to the current app and url name
+    (request.resolver_match.app_name and request.resolver_match.url_name).
+    """
+
+    permission_required = 'projectroles.view_project'
+
+    def get(self, request, *args, **kwargs):
+        project = self.get_project()
+        app_name = request.GET.get('app_name')
+        url_name = request.GET.get('url_name')
```
mikkonie: As discussed earlier, we can simply omit `url_name` here. By providing the app_name, the client side app can provide the correct app to be highlighted regardless of what page we are.

This will be mostly be used by single-page apps which don't have a concept of multiple separate Django URLs anyway. The URL name is used for `projectroles` URLs since there are multiple UI elements highlighted depending on which projectroles view we are on. These Ajax views are called by other apps.

projectroles/views_ajax.py
```diff
@@ -431,6 +431,51 @@ def post(self, request, *args, **kwargs):
...
+    attribute for each of the links, you must provide the app_name and url_name
+    as GET parameters. Both of them refer to the current app and url name
+    (request.resolver_match.app_name and request.resolver_match.url_name).
+    """
+
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request, *args, **kwargs):
+        app_name = request.GET.get('app_name')
+        url_name = request.GET.get('url_name')
```
mikkonie: Same thing here, we can simply omit `url_name` from the Ajax views.

mikkonie (changes requested):
Good work! I added some requests to remove the `url_name` parameters from Ajax views as they are not necessary.

projectroles/tests/test_views_ajax.py
```diff
@@ -461,6 +461,431 @@ def test_project_star(self):
...
+                    'url': f'/project/project/update/{self.project.sodar_uuid}',
+                    'label': 'Update Project',
+                    'icon': 'mdi:lead-pencil',
+                    'active': False,
+                },
+            ],
+        }
+        self.assertEqual(response.json(), expected)
+
+    def test_get_no_access(self):
```
mikkonie, line 615: Permission tests should go in `projectroles.tests.test_permissions_ajax`. Please add permission tests for both views, follow the conventions in that test module and, needless to say, ask me if anything is unclear.

mikkonie (changes requested):
Looking good! One more comment regarding permission tests for the ajax views. Once we have permission tests in place, I'll be happy to merge.

projectroles/tests/test_permissions_ajax.py
```diff
@@ -237,6 +236,101 @@ def test_get_category_anon(self):
...
+            self.user_delegate_cat,
+            self.user_contributor_cat,
+            self.user_guest_cat,
+            self.user_owner,
+            self.user_delegate,
+            self.user_contributor,
+            self.user_guest,
+        ]
+        bad_users = [self.user_no_roles, self.user_finder_cat, self.anonymous]
+        self.assert_response(self.url, good_users, 200, method='GET')
```
mikkonie: Since `GET` is the default value for `method` here, we can just omit the parameter here for a bit more compact tests.

projectroles/tests/test_permissions_ajax.py
```diff
@@ -237,6 +236,101 @@ def test_get_category_anon(self):
...
+            self.user_owner_cat,
+            self.user_delegate_cat,
+            self.user_contributor_cat,
+            self.user_guest_cat,
+            self.user_owner,
+            self.user_delegate,
+            self.user_contributor,
+            self.user_guest,
+        ]
+        bad_users = [self.user_no_roles, self.user_finder_cat, self.anonymous]
```
mikkonie: Let's follow the conventions for ordering the users for better readability: first all the users with roles, then user_no_roles, ending with anonymous.

projectroles/tests/test_permissions_ajax.py
```diff
@@ -237,6 +236,101 @@ def test_get_category_anon(self):
...
+        self.assert_response(self.url, good_users, 200, method='GET')
+        self.assert_response(self.url, bad_users, 403, method='GET')
+        self.project.set_public()
+        self.assert_response(self.url, self.user_no_roles, 200, method='GET')
```
mikkonie: You should add a test case for when the project is archived. See elsewhere in tests for an example. Note that this is only required for the project, as categories can not be archived.

mikkonie (changes requested):
Some cleanup comments, otherwise looking good.

mikkonie (approved):

PR #1422, opened 2024-04-29
Form proper labels for Category/Project update link

### LINKED ISSUE #1421: Incorrect project type strings returned for sidebar and project dropdown
### Describe the Bug

Something we missed in #1380 it seems: the correct project type is no longer used in the sidebar links. When viewing the links in a category, the update view link is displayed as `Update Project`.

### How to Reproduce

Create a category, navigate into the category and observe the `Update Project` link in the sidebar instead of `Update Category`.

### Expected Behavior

In case of a category, the link should display `Update Category`. For a project, it should say `Update Project`.

Furthermore, we should be using `get_display_name()` to return the project and category labels as configured by the site using SODAR Core. In some sites based on this platform, different terms are used for projects and categories.

All this logic was correctly present in the original implementation in the templates, but for some reason hardcoded strings are currently used in `AppLinkContent.get_project_app_links()`. The original behaviour should be replicated.

Tests will of course need to be updated to catch and assert this correct behaviour.

### Screenshots

2024-04_sidebar_project_type_bug

gromdimon:
Issue: #1421

### DIFF
--- a/projectroles/tests/test_templatetags.py
+++ b/projectroles/tests/test_templatetags.py
@@ -34,7 +34,6 @@ from projectroles.tests.test_models import (
 )
 
 
-
 app_settings = AppSettingAPI()
 site = import_module(settings.SITE_PACKAGE)
 
--- a/projectroles/tests/test_templatetags.py
+++ b/projectroles/tests/test_templatetags.py
@@ -689,7 +689,7 @@ class TestProjectrolesTags(TemplateTagTestBase):
                 },
                 {
                     'name': 'project-update',
-                    'url': f'/project/project/update/{self.project.sodar_uuid}',
+                    'url': f'/project/update/{self.project.sodar_uuid}',
                     'label': 'Update Project',
                     'icon': 'mdi:lead-pencil',
                     'active': False,
--- a/projectroles/tests/test_templatetags.py
+++ b/projectroles/tests/test_templatetags.py
@@ -689,7 +689,7 @@ class TestProjectrolesTags(TemplateTagTestBase):
                 },
                 {
                     'name': 'project-update',
-                    'url': f'/project/update/{self.project.sodar_uuid}',
+                    'url': f'/project/project/update/{self.project.sodar_uuid}',
                     'label': 'Update Project',
                     'icon': 'mdi:lead-pencil',
                     'active': False,
--- a/projectroles/tests/test_templatetags.py
+++ b/projectroles/tests/test_templatetags.py
@@ -34,6 +34,7 @@ from projectroles.tests.test_models import (
 )
 
 
+
 app_settings = AppSettingAPI()
 site = import_module(settings.SITE_PACKAGE)
 
--- a/projectroles/utils.py
+++ b/projectroles/utils.py
@@ -183,6 +183,7 @@ class AppLinkContent:
         ret = []
         # Add project related links
         if project:
+            project_display_name = get_display_name(project.type, title=True)
             # Add project overview link
             ret.append(
                 {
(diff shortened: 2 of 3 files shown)


projectroles/utils.py
```diff
@@ -191,7 +191,7 @@ def get_project_app_links(
                         'projectroles:detail',
                         kwargs={'project': project.sodar_uuid},
                     ),
-                    'label': 'Project Overview',
+                    'label': 'Overview',
```
mikkonie: This one is just "Overview" now, the project/category title is missing.

projectroles/utils.py
```diff
@@ -249,7 +249,7 @@ def get_project_app_links(
                             'projectroles:update',
                             kwargs={'project': project.sodar_uuid},
                         ),
-                        'label': 'Update Project',
+                        'label': f'Update {get_display_name(project.type)}',
```
mikkonie: You need to supply `title=True` as in the original implementation to ensure the name is returned capitalized. This was present in the original implementation. Please test these features to ensure they appear correct before submitting a pr.

projectroles/utils.py
```diff
@@ -249,7 +249,7 @@ def get_project_app_links(
...
                         'icon': 'mdi:lead-pencil',
                         'active': self._is_active_projectroles(
                             link_names=['update'],
```
mikkonie, line 256: For all the links after this point in the code, all the labels are still provided as hardcoded text. Please go through all the changes required carefully before submitting.

projectroles/tests/test_templatetags.py
```diff
@@ -690,7 +691,7 @@ def test_get_sidebar_links_project_detail_view(self):
                 {
                     'name': 'project-update',
                     'url': f'/project/project/update/{self.project.sodar_uuid}',
-                    'label': 'Update Project',
+                    'label': f'Update {get_display_name(self.project.type)}',
```
mikkonie: Probably better to check for a literal string here?

mikkonie (changes requested):

projectroles/utils.py
gromdimon: I'm sorry, that I overlooked it! Now it should be fixed

gromdimon:
It looks like some 'label' lines are now too long. Although Black and Flake8 do not raise any issues, this definitely doesn't follow conventions. There are multiple ways to solve this problem: reformatting the string using parentheses (), using the '+' syntax, adding a helper method, or introducing some temporary variables before appending the link.

The question now is: how should I address this issue?

mikkonie:
> It looks like some 'label' lines are now too long. Although Black and Flake8 do not raise any issues, this definitely doesn't follow conventions. There are multiple ways to solve this problem: reformatting the string using parentheses (), using the '+' syntax, adding a helper method, or introducing some temporary variables before appending the link.
> 
> The question now is: how should I address this issue?

This is a good point. You have "wrapped" the code inside the format string and strings are exempt from line length formatting in black. It is possible to do the following:

```
'label': f'Create '
'{get_display_name(PROJECT_TYPE_PROJECT, title=True)}',
```

That reduces line length and is accepted by black. Probably the best way unless you want to use the old style `format()` or get the title into a variable before inserting it into the string.

projectroles/utils.py
```diff
@@ -191,7 +191,7 @@ def get_project_app_links(
                         'projectroles:detail',
                         kwargs={'project': project.sodar_uuid},
                     ),
-                    'label': 'Project Overview',
+                    'label': f'{get_display_name(project.type, title=True)} Overview',
```
mikkonie: We could get the string for the current project's type into a variable and use it both here and in the "Update Project" link. That would save us from calling the same method twice with identical input.

mikkonie (changes requested):
I had one minor comment, plus I gave a hint on how to fix the line length issue in the pull request comments. Otherwise this looks fine and once these small things are fixed, I'll merge.

mikkonie (approved):

PR #1423, opened 2024-04-29
Remove `project/` prefix from urls

### LINKED ISSUE #1392: Incorrect URL routes for project create/update/archive views
I noticed we have the following routes in `projectroles.urls`:

```
route='project/update/<uuid:project>'
route='project/create'
route='project/create/<uuid:project>'
route='project/archive/<uuid:project>'
```

This is wrong, as the `project/` prefix is already provided by including the `projectroles` app URLs into the site URL config. This creates URLs with `project/project/x`. It seems we have had this bug in since v0.13.0 but I've missed it. 

The `project/` prefix should be dropped from these routes.

Since we use URL names in our code everywhere, this should not break anything (except for possible user bookmarks for the erroneous URLs).

gromdimon:
Issue: #1392 

I just removed the prefixes and everything seems to be working :)

### DIFF
--- a/projectroles/tests/test_templatetags.py
+++ b/projectroles/tests/test_templatetags.py
@@ -689,7 +689,7 @@ class TestProjectrolesTags(TemplateTagTestBase):
                 },
                 {
                     'name': 'project-update',
-                    'url': f'/project/project/update/{self.project.sodar_uuid}',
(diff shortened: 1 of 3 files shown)


mikkonie (approved):
Looks good, approved.

mikkonie:
Done by @gromdimon, merged.

PR #1424, opened 2024-04-29
Use constants for timeline event status types

### LINKED ISSUE #973: Use constants for timeline event status types
Similar to bihealth/sodar-server#1398, we only use strings here. It's even worse in this case since this is an app designed to be used by other apps.

Should use constants instead. However, since `TimelineAPI` is used as a dynamic backend plugin API class, we should not have to import them directly.

## Spec

- Define constants for status types in `timeline.models`
    * `TL_STATUS_OK = 'OK'` etc
    * Also set these in `EVENT_STATUS_TYPES` and `DEFAULT_MESSAGES`
- Import constants to `timeline.api`, set them as attributes of the `TimelineAPI` class
    * `TL_STATUS_OK = TL_STATUS_OK` etc
    * No need to use `__init__()`, just set them directly in the class
- Update usage of status types from strings to constants everywhere in the project
    * E.g. `timeline.add_event(status_type=timeline.TL_STATUS_OK)` etc
    * No direct imports anywhere outside the timeline app!
    * Also remember to update this in all tests
- Also update usage inside the timeline app whereever pure strings are used now
- Update `Defining Status States` section in `app_timeline_usage` doc
    * Also anywhere else where these may appear in the tocs

## Tasks

- [x] Add constants
- [x] Set constants as `TimelineAPI` class attributes
- [x] Update usage
- [x] Update tests
- [x] Update docs

gromdimon:
Issue: #973

### DIFF
--- a/bgjobs/views.py
+++ b/bgjobs/views.py
@@ -100,7 +100,7 @@ class BackgroundJobClearViewBase(
                     description='Clearing {} background jobs'.format(
                         'user-owned' if self.which_jobs != 'all' else 'all'
                     ),
-                    status_type='OK',
+                    status_type=timeline.TL_STATUS_OK,
                 )
             messages.success(
                 self.request, 'Removed {} background jobs.'.format(bg_job_count)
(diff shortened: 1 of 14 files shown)


mikkonie (approved):
Looking good! No notes, will merge.

Issue #1513, opened 2024-11-18
Add general purpose API exception class for HTTP 503


mikkonie:
An exception class for `HTTP 503` is not included in DRF but is quite commonly needed in SODAR. It would be nice to provide this as a general purpose class.

Alternatively: look for a DRF extension which provides extended out of the box. Maybe one exists?

Naturally, after doing this we should replace any locally created exception classes with the generic one. In SODAR Core, such a local class only exists in `SodarcacheAPIViewMixin.BackendUnavailable`.

## Spec

- Create generic HTTP 503 exception class
    * ..or, add dependency to some well-maintained DRF extension if one exists?
    * If adding a new one, place this in `projectroles.views_api`
- Replace locally defined exception class in `SodarcacheAPIViewMixin`
- Ensure CI passes (no new tests required)
- Document
    * If we add a new custom exception class of our own, we should to mention it in `dev_project_app.rst` under "REST API Development"

## Tasks

- [ ] Create or import a generic 503 exception class
- [ ] Replace local implementation
- [ ] Document

PR #1864, opened 2026-01-16
Add general purpose API exception class for HTTP 503

fmarotta:
This PR attempts to solve #1513 by implementing the specification described therein.

### DIFF
--- a/docs/source/dev_project_app.rst
+++ b/docs/source/dev_project_app.rst
@@ -769,7 +769,7 @@ The base classes provide permission checks via SODAR Core project objects
 similar to UI view mixins. Base REST API classes without a project context can
 also be used in site apps.
 
-For views which returns ``HTTP 503`` errors, we provide a general purpose
+For views which return ``HTTP 503`` errors, we provide a general purpose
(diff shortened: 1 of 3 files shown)


docs/source/dev_project_app.rst
```diff
@@ -769,6 +769,9 @@ The base classes provide permission checks via SODAR Core project objects
 similar to UI view mixins. Base REST API classes without a project context can
 also be used in site apps.
 
+For views which returns ``HTTP 503`` errors, we provide a general purpose
```
mikkonie: I noticed a small typo here, should be "return" instead of "returns". Other than that, looks good to me :)

mikkonie (changes requested):
One small fix for documentation requested, otherwise looking good!

mikkonie (approved):

docs/source/dev_project_app.rst
fmarotta: Thanks for the review! It's fixed now. Should I stop one of the useless CI pipelines? (I also accidentally triggered one with a silly formatting commit)

Issue #1853, opened 2025-12-16
Add optional display of username in user dropdown button


mikkonie:
We should provide an option to display the `username` of the current user directly in the user dropdown button. This should be done with a `USER` scope app setting.

As this feature is mostly usable for power users with multiple accounts (with e.g. multi-organization LDAP access and/or users of service accounts) and in development, it should be disabled by default.

## Spec

- Define new `PluginAppSettingDef` in `projectroles`
    * name: `'dropdown_user_name_display'`
    * scope: `APP_SETTING_SCOPE_USER`
    * type: `APP_SETTING_TYPE_BOOLEAN`
    * default: `False`
    * label: `'Display user name in user dropdown'`
    * description: `'Display user name of current user in the user dropdown button'`
    * user_modifiable: `True`
    * global_edit: `True`
- Add display logic in `_site_titlebar_dropdown.html`
    * If we have an authenticated user, retrieve setting using the `get_app_setting()` template tag
        + (In the future we might want to set things like this in a context processor as it performs better, but that's for later. See #1604)
    * If setting is `True`, display `request.user.username` in addition to the user icon
    * **TBD:** Display as `badge-pill` if setting is enabled?
        + This would probably look better than simply inserting the username as text
        + Feel free to experiment on what looks good
        + **Hint:** Whatever design you use, it's a good idea to add an `id` attribute to the element to make UI tests easier :)
    * **TBD:** Truncate very long user names?
        + Again, feel free to experiment. There might be some cases (e.g. OIDC user preferred username check failing) where we end with a very long username
- Add UI tests
    * These should probably go into `projectroles.tests.test_ui.TestBaseTemplate`
    * Test case 1: Test visibility for user with setting default (`False`), username should not be visible
    * Test case 2: Test visibility for user with setting set `True`, ensure user name is visible
    * Test case 3: Test without authenticated user and `PROJECTROLES_ALLOW_ANONYMOUS=True`
        + Username should not be visible, sanity check to ensure view still renders correctly
    * (+ possible test for truncation if implemented, any other relevant things to test I missed?)
- Document
    * Add description of setting in `app_userprofile.rst` under `User Settings`

## Tasks

- [ ] Add app setting
- [ ] Add display logic in template
- [ ] Add UI tests
- [ ] Update documentation

PR #1866, opened 2026-01-20
Add optional display of username in user dropdown button

fmarotta:
This PR aims to solve #1853 by implementing the spec described therein. The username is displayed in a badge-pill and is truncated if very long.


fmarotta:
Update on the design: currently it looks like this
Screenshot from 2026-01-21 14-05-32
Pros:
- The colors of both icon and pill are updated consistently upon hover and click

Cons:
- The colors are hard-coded in the CSS, as I couldn't find a suitable bootstrap class

---

I could also envision the username as a plain text element maybe, which would look like so
Screenshot from 2026-01-21 13-27-36
Pros:
- Simpler implementation with no hard-coded colors
- Style would be consistent with the adjacent elements in the navbar (i.e., icon + plain text)

---

At this point I'm not sure which one is better, I am happy to hear if you have a preference or other suggestions

projectroles/app_settings.py
```diff
@@ -232,6 +232,16 @@
         global_edit=False,
         project_types=[PROJECT_TYPE_CATEGORY],
     ),
+    PluginAppSettingDef(
+        name='dropdown_user_name_display',
+        scope=APP_SETTING_SCOPE_USER,
+        type=APP_SETTING_TYPE_BOOLEAN,
+        default=False,
+        label='Display user name in user dropdown',
+        description='Display user name of current user in the user dropdown button',
```
mikkonie: This is done exactly as I specified it, that's good :)

However, looking at it in the browser, maybe it's better to write this as "Display your user name in the user dropdown link".

projectroles/static/projectroles/css/projectroles.css
```diff
@@ -594,6 +594,21 @@ a.sodar-pr-btn-copy-secret i {
   50% { opacity: 0 }
 }
 
+/* User dropdown username */
+#sodar-navbar-user-dropdown-username {
+  background-color: rgba(255,255,255,.5);
```
mikkonie: Since we are using Bootstrap, I think we should simply use `badge-secondary` instead of defining foreground and background manually in the CSS. Unless there is a reason using the Bootstrap variants does not work here?

Other parameters such as maximum width and alignment are fine to enter as CSS as fine-tuned control would require us doing our own theme.

Also, the hover highlighting looks very good so if this is how to achieve that, please keep it as is :)

projectroles/templates/projectroles/_site_titlebar_dropdown.html
```diff
@@ -35,7 +35,14 @@
     {% if request.user.is_superuser %}
       <i class="iconify" data-icon="mdi:shield-account"></i>
     {% elif request.user.is_authenticated %}
+      {% get_app_setting 'projectroles' 'dropdown_user_name_display' user=request.user as dropdown_user_name_display %}
       <i class="iconify" data-icon="mdi:user"></i>
```
mikkonie: I would prefer having the user icon within the badge. This would be consistent with other UI elements such as project and user badges in the timeline.

projectroles/templates/projectroles/_site_titlebar_dropdown.html
```diff
@@ -35,7 +35,14 @@
     {% if request.user.is_superuser %}
       <i class="iconify" data-icon="mdi:shield-account"></i>
```
mikkonie: This is probably the fault of how I wrote doen the specification: This feature should be applied to all authenticated users, including superusers :) Right now, superusers don't get the name in the dropdown even if they enable the setting.

projectroles/tests/test_ui.py
```diff
@@ -112,6 +112,77 @@ def test_admin_link_modal(self):
         )
         self.assertEqual(modal.is_displayed(), True)
 
+    def test_user_dropdown_username_default(self):
+        """Test visibility for user with setting default (False)"""
+        # username should not be visible
```
mikkonie: Sorry for minor nitpicking, but let's start comments with a capital letter. That is consistent with conventions and helps readability in case of e.g. multiline comments.

projectroles/tests/test_ui.py
```diff
@@ -112,6 +112,77 @@ def test_admin_link_modal(self):
...
+        with self.assertRaises(NoSuchElementException):
+            self.selenium.find_element(
+                By.ID, 'sodar-navbar-user-dropdown-username'
+            )
+
+    def test_user_dropdown_username_superuser(self):
+        """Test visibility with superuser"""
+        # username should not be visible
+        self.login_and_redirect(self.superuser, reverse('home'))
+        with self.assertRaises(NoSuchElementException):
```
mikkonie: This should of course be updated with expecting the badge to be displayed if we are a superuser.

projectroles/tests/test_ui.py
```diff
@@ -112,6 +112,77 @@ def test_admin_link_modal(self):
...
+            APP_NAME, 'dropdown_user_name_display', True, user=long_name_user
+        )
+        self.login_and_redirect(long_name_user, reverse('home'))
+        username_element = self.selenium.find_element(
+            By.ID, 'sodar-navbar-user-dropdown-username'
+        )
+        # Based on https://medium.com/@dave_go/testing-text-truncation-effective-strategies-2a589841a22b
+        # We compare offsetWidth (which should be equal to the max-width css
+        # property) to scrollWidth (which is the total width)
+        self.assertTrue(
```
mikkonie: We could also simply use the Django `truncatechars` template tag in the template, e.g. `{{ request.user.username|truncatechars:32 }}`. I'll let you choose whcih way is better :)

https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#truncatechars

mikkonie (changes requested):
Good work! I found some minor things to tweak as well as one "bug", but I guess that one is more of a problem with how I wrote the specification.

projectroles/app_settings.py
fmarotta: Done. Actually I hadn't properly checked the settings page in the browser. I have also moved the option up in the list, before "project_list_pagination" (I think it looks a bit better if it's shown together with the other boolean options), but I'm happy to move it back in last place if you want.

projectroles/static/projectroles/css/projectroles.css
fmarotta: Indeed, I was not very happy about setting the colors manually, but I experienced the following issues:
- With `badge-secondary`, the badge would not have the same color as the icon (it would be a darker shade of gray, which doesn't work well with in combination with the hover color and the black foreground color).
- The bootstrap classes used for the icon and the text of the other elements in the title bar change the foreground color, whereas for the badge we need to change the background color.

Basically, I couldn't find any bootstrap class that changes the background color of the badge to be the same as the foreground color of the icon and the other elements. Such foreground color (including the hover higlighting) is achieved by the class combo `navbar-dark .navbar-nav .nav-link`. However, this combo cannot be used to change the background color of the badge.

As far as I can tell, there may be one way to avoid defining foreground and background colors manually, but it involves changing the classes of the other elements in the title bar (I mean the "Extra Link" and the user icon).

projectroles/templates/projectroles/_site_titlebar_dropdown.html
fmarotta: Done

projectroles/templates/projectroles/_site_titlebar_dropdown.html
fmarotta: Done, sorry I misinterpreted the spec. I've put the logic in a separate file to avoid a bit of duplication, but please let me know if `{% include %}` shouldn't be used for such things.

projectroles/tests/test_ui.py
fmarotta: Done

projectroles/tests/test_ui.py
fmarotta: :+1:

projectroles/tests/test_ui.py
fmarotta: I accept the suggestion, it's more straightforward this way :)

fmarotta:
Thank you for the comments, I have addressed them and the PR should be ready for another round of review, once the CI passes.

projectroles/static/projectroles/css/projectroles.css
fmarotta: On second thought, maybe there is another way which I had previously missed: using an `<a>` element for the badge. In that case, hovering should work with bootstrap classes. Let me give it a try.

fmarotta: OK in that case it could look something like this:
image
with the following snippet:
```html
<li class="nav-item sodar-navbar-user-dropdown">⏎                                                                                                                
  <a class="nav-link badge-pill py-1 my-1 ml-1 badge-secondary dropdown-toggle"⏎                                                                                 
     id="sodar-navbar-user-dropdown"⏎                                                                                                                            
     data-toggle="dropdown">⏎                                                                                                                                    
    <i class="iconify" data-icon="mdi:shield-account"></i>⏎                                                                                                      
    {{ request.user.username|truncatechars:32 }}⏎                                                                                                                
  </a>
```

fmarotta:
CI had failed because I didn't realize I had to define the setting `dropdown_user_name_display` also in another test file: in that test, the superuser was logged in and the template was expecting to find this setting in order to decide whether to render the username in the dropdown or not. Should be fixed now.

projectroles/static/projectroles/css/projectroles.css
mikkonie: Fair enough. This looks good and it's not too much extra CSS, so this seems like a good approach :)

projectroles/templates/projectroles/_site_titlebar_dropdown.html
```diff
@@ -33,9 +33,9 @@
      id="sodar-navbar-user-dropdown"
      data-toggle="dropdown">
     {% if request.user.is_superuser %}
-      <i class="iconify" data-icon="mdi:shield-account"></i>
+      {% include "projectroles/_site_titlebar_user_dropdown_link.html" with icon="mdi:shield-account" %}
     {% elif request.user.is_authenticated %}
-      <i class="iconify" data-icon="mdi:user"></i>
+      {% include "projectroles/_site_titlebar_user_dropdown_link.html" with icon="mdi:user" %}
```
mikkonie: This works and it's fine, but we could optimize away some repetition. I suggest the following:

1. Add a `get_user_icon()` template tag under `projectroles_common_tags`
   * Receives user object as argument, returns icon string with namespace based on superuser/authentication/anonymous status
   * Why common tags? Because this could be useful elsewhere too..
   * Remember to also test this in test_templatetags
2. Call the new template tag here and store the icon as variable
   * E.g. `{% get_user_icon request.user as user_icon %}`
3. Move the app setting check and resulting logic back into `_site_titlebar_dropdown.html`
   * Refactor it to remove repetition:
       + IF dropdown_user_name_display AND user.is_superuser OR user is_authenticated THEN display badge
       + ELSE simply display icon
4. Remove the include template as it is no longer needed

Sound good?

mikkonie (changes requested):
Good work! Functionally this looks perfect, nice tests too. I still have one suggestion for refactoring.

projectroles/templates/projectroles/_site_titlebar_dropdown.html
fmarotta: Very good, I have done as suggested. There were already some common tags related to user icons, maybe we could consider some refactoring in the future.

mikkonie: Great, thank you.

You are right, we might want to unify the icon template tags into one tag which could either return just the tag ID or full HTML like `get_user_superuser_icon()` and `get_user_inactive_icon()` currently do.

This being the common template tag class, such a change should be done by deprecation: first deprecate the old implementations in a major release, then remove completely in the next one.

Feel free to open an issue for that one, we can then refined that and decide on prioritization and milestone.

projectroles/templates/projectroles/_site_titlebar_dropdown.html
```diff
@@ -29,17 +29,24 @@
...
-    {% elif request.user.is_authenticated %}
-      <i class="iconify" data-icon="mdi:user"></i>
-    {% else %}
-      <i class="iconify" data-icon="mdi:incognito"></i>
-    {% endif %}
-  </a>
+  {% get_app_setting 'projectroles' 'dropdown_user_name_display' user=request.user as dropdown_user_name_display %}
+  {% get_user_icon request.user as user_icon %}
+  {% if dropdown_user_name_display and request.user.is_authenticated %}
+    <a class="nav-link badge-pill badge-secondary py-1 my-1 ml-1 dropdown-toggle"
```
mikkonie: One problem I noticed.. The prior version I reviewed had a very nice layout for the dropdown:

Image

However, currently it looks like this:

Image

Maybe this is due to the [removal of CSS here](https://github.com/bihealth/sodar-core/pull/1866/changes/4a00b4956f3de65e9675a966410fb092dc135c7d#diff-79af83a01e03880cf409ff9693a709c9292b10b896ec210277d402ef28e68c97L597)?

projectroles/tests/test_ui.py
```diff
@@ -10,6 +10,7 @@
 from django.test import override_settings
 from django.urls import reverse
 from django.utils import timezone
+from django.template.defaultfilters import truncatechars
```
mikkonie: One super small cleanup thing: let's include these imports in alphabetical order for maximum readability :) so this should go before `django.test`.

mikkonie (changes requested):
I noticed one bug (?) and one very minor cleanup thing. Good work otherwise, once these are fixed I'm good to merge.

projectroles/tests/test_ui.py
fmarotta: Good catch, I missed that. Since we already use `black`, would you be open to adding `isort`? It has many features and supports [custom sections](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html), so we should be able to preserve the groups defined [in the docs](https://sodar-core.readthedocs.io/en/latest/dev_core_guide.html#module-imports).

projectroles/templates/projectroles/_site_titlebar_dropdown.html
fmarotta: Indeed. I have now reverted the layout to the former version.

mikkonie (approved):
No notes, looks and works great. Merging :)

Issue #1434, opened 2024-06-18
Use UniqueConstraint objects instead of unique_together


mikkonie:
[Django documentation hints that](https://docs.djangoproject.com/en/5.2/ref/models/options/#unique-together) `unique_together` will be deprecated at some point in the future. Instead, we should use `constraints` and `UniqueConstraint` objects [as documented here](https://docs.djangoproject.com/en/5.2/ref/models/constraints/#django.db.models.UniqueConstraint).

## Spec

- Replace `unique_together` constraints with `UniqueConstraint`
    * Preserve fields as they are currently defined (of course)
    * See Django v5.2 docs for more info
- Run `./manage.py makemigrations` if needed
    * Probably are needed, but I'm not sure?
- Ensure CI passes

mikkonie:
Done and merged, closing.

PR #1867, opened 2026-01-20
Use UniqueConstraint instead of unique_together

fmarotta:
With regard to issue #1434, this PR replaces the use of `unique_together` with more powerful `UniqueConstraint` objects.


mikkonie (approved):
No notes, looking good :)

Issue #1550, opened 2025-01-22
Add host name deletion confirmation element as reusable helper


mikkonie:
In two separate places, we have a UI `DeleteView` which requires inputting the host name of the server before deleting some sensitive data.

The views in question are:
- `ProjectDeleteView` in SODAR Core
    * [View](https://github.com/bihealth/sodar-core/blob/dev/projectroles/views.py#L1973C7-L1973C24)
    * [Template](https://github.com/bihealth/sodar-core/blob/dev/projectroles/templates/projectroles/project_confirm_delete.html)
- `SheetDeleteView` in SODAR Server
    * [View](https://github.com/bihealth/sodar-server/blob/dev/samplesheets/views.py#L1754)
    * [Template](https://github.com/bihealth/sodar-server/blob/dev/samplesheets/templates/samplesheets/samplesheet_confirm_delete.html)

As can be seen, there's quite a bit of repeated code between these views and forms. Instead, I would like to provide this extension to `DeleteView` as a reusable helper in SODAR Core.  

The way I see it, there are two ways to do this:

- Provide a view extending Django's `DeleteView` which includes a form with the host name widget and validation
- OR, just provide a common template tag for widget, along with a mixin for validation logic. These could then be embedded into standard `DeleteView` views.

Feel free to look into which method of implementation would be nicer. The overall goal is to:

- Provide host name widget in delete form (in the projectroles app)
- Provide host name input validation for the form (in the projectroles app)
- Maintain current functionality of the existing views with as little of repeated code as possible

The views already have tests for correct and invalid host name input, so no new tests should be required.

If we add new base classes, mixins and/or template tags, they should be documented. This is done by adding proper docstrings and then using Sphinx autodoc. See `app_projectroles_api_django.rst`.

## Tasks

- [ ] Replace repeated code with a reusable implementation
- [ ] Use reusable implementation in `ProjectDeleteView`
- [ ] Document

PR #1871, opened 2026-01-26
Provide GenericDeleteView with host name confirmation (#1550)

fmarotta:
This PR adds a `GenericDeleteView` class, addressing #1550, and uses it to implement `ProjectDeleteView`, maintaining the existing logic for access control and redirection in various cases.

### DIFF
--- a/docs/source/app_projectroles_api_django.rst
+++ b/docs/source/app_projectroles_api_django.rst
@@ -35,6 +35,10 @@ Base mixins and classes for building UI views can be found in
     :members:
     :show-inheritance:
 
+.. autoclass:: GenericDeleteView
+    :members:
+    :show-inheritance:
+
 .. autoclass:: ProjectContextMixin
     :members:
     :show-inheritance:
--- a/docs/source/app_projectroles_api_django.rst
+++ b/docs/source/app_projectroles_api_django.rst
@@ -35,10 +35,6 @@ Base mixins and classes for building UI views can be found in
     :members:
     :show-inheritance:
 
-.. autoclass:: HostConfirmDeleteView
-    :members:
-    :show-inheritance:
-
 .. autoclass:: ProjectContextMixin
     :members:
     :show-inheritance:
@@ -48,6 +44,10 @@ Base mixins and classes for building UI views can be found in
     :show-inheritance:
 
 .. autoclass:: InvalidFormMixin
+    :members:
+    :show-inheritance:
+
+.. autoclass:: HostConfirmDeleteView
     :members:
     :show-inheritance:
 
--- a/docs/source/app_projectroles_api_django.rst
+++ b/docs/source/app_projectroles_api_django.rst
@@ -35,7 +35,7 @@ Base mixins and classes for building UI views can be found in
     :members:
     :show-inheritance:
 
-.. autoclass:: GenericDeleteView
+.. autoclass:: HostConfirmDeleteView
     :members:
     :show-inheritance:
 
--- a/projectroles/templates/projectroles/confirm_delete_host.html
+++ b/projectroles/templates/projectroles/confirm_delete_host.html
@@ -12,17 +12,9 @@
   <h3>Confirm Deletion of {{ type_display_name|title }}</h3>
 </div>
 <div class="container-fluid sodar-page-container">
-  {% if warning_alert_message %}
-    <div class="alert alert-warning" role="alert">
-      {{ warning_alert_message|safe }}
-    </div>
-  {% endif %}
-
-  {% if danger_alert_message %}
-    <div class="alert alert-danger" role="alert">
-      {{ danger_alert_message|safe }}
-    </div>
-  {% endif %}
+  {% block custom_alerts %}
+    {% include custom_alerts_include %}
+  {% endblock custom_alerts %}
 
   <form method="post">
     {% csrf_token %}
@@ -33,7 +25,7 @@
            placeholder="Type the host name of this server (e.g. &quot;site.example.com&quot;) to confirm deletion." />
     <div class="btn-group pull-right">
       <a role="button" class="btn btn-secondary"
-          href="{{ redirect_url }}">
+         href="{{ request.session.real_referer }}">
         <i class="iconify" data-icon="mdi:arrow-left-circle"></i> Cancel
       </a>
       <button type="submit" class="btn btn-danger sodar-btn-submit-once"
--- a/projectroles/templates/projectroles/confirm_delete_host.html
+++ b/projectroles/templates/projectroles/confirm_delete_host.html
@@ -3,18 +3,23 @@
 {% load rules %}
 
 {% block title %}
-  Confirm Deletion of {{ type_display_name|title }}
+  Confirm Deletion of {{ object_display_name|title }}
 {% endblock title %}
 
 {% block projectroles_extend %}
 
 <div class="row sodar-subtitle-container">
-  <h3>Confirm Deletion of {{ type_display_name|title }}</h3>
+  {% block form_header %}
+    <h3>Confirm Deletion of {{ object_display_name|title }}<h3>
+  {% endblock form_header %}
 </div>
 <div class="container-fluid sodar-page-container">
-  {% block custom_alerts %}
-    {% include custom_alerts_include %}
-  {% endblock custom_alerts %}
+  {% block custom_content %}
+    <div class="alert alert-danger" role="alert">
+      Are you sure you want to delete this object?
+      This action can <strong>NOT</strong> be undone!
+    </div>
+  {% endblock custom_content %}
 
   <form method="post">
     {% csrf_token %}
--- a/projectroles/views.py
+++ b/projectroles/views.py
@@ -2016,12 +2016,12 @@ class ProjectArchiveView(
         return redirect(redirect_url)
 
 
-class GenericDeleteView(
+class HostConfirmDeleteView(
     LoginRequiredMixin,
     LoggedInPermissionMixin,
     DeleteView,
 ):
-    """Generic deletion view with confirmation form"""
+    """Specialized deletion view with confirmation form"""
 
     template_name = 'projectroles/delete_confirmation_form.html'
 
@@ -2091,7 +2091,7 @@ class ProjectDeleteView(
     ProjectDeleteMixin,
     ProjectDeleteAccessMixin,
     ProjectModifyPluginViewMixin,
-    GenericDeleteView,
+    HostConfirmDeleteView,
 ):
     """Project deletion view"""
 
--- a/projectroles/views.py
+++ b/projectroles/views.py
@@ -2071,7 +2071,7 @@ class GenericDeleteView(
             return self.on_error(
                 object,
                 self.request,
-                'Host name input incorrect, deletion cancelled.',
+                'Host name input incorrect.',
                 self.get_redirect_url(object),
             )
 
--- a/projectroles/views.py
+++ b/projectroles/views.py
@@ -1672,14 +1672,60 @@ class ProjectDeleteMixin(ProjectModifyPluginViewMixin):
         )
 
     @classmethod
-    def get_redirect_url(cls, project: Project) -> str:
+    def on_success(cls, project: Project, request: HttpRequest):
+        """
+        Called when project deletion is successful. It should print
+        a status message and redirect the user.
+
+        :param project: Project object
+        :param request: HttpRequest object
+        """
+        p_type = get_display_name(project.type, title=True)
+        messages.success(request, f'{p_type} deleted.')
         if project.parent:
-            return reverse(
+            redirect_url = reverse(
                 'projectroles:detail',
                 kwargs={'project': project.parent.sodar_uuid},
             )
         else:
-            return reverse('home')
+            redirect_url = reverse('home')
+        return redirect(redirect_url)
+
+    @classmethod
+    def on_error(
+        cls,
+        project: Project,
+        request: HttpRequest,
+        err: str,
+        redirect_url: Optional[str] = None,
+    ):
+        """
+        Called when project deletion fails. It should print an error message
+        and redirect the user.
+
+        :param project: Project object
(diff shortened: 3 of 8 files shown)


projectroles/views.py
```diff
@@ -1970,38 +2016,48 @@ def post(self, request, **kwargs):
         return redirect(redirect_url)
 
 
-class ProjectDeleteView(
+class GenericDeleteView(
```
mikkonie: This is in fact intended as a specialized version of the "generic" Django `DeleteView`, so I think we should name it more descriptively. I recommend `HostConfirmDeleteView`. That would also be in line with general naming conventions.

projectroles/views.py
```diff
@@ -1672,14 +1672,60 @@ def _create_timeline_event(cls, project: Project, request: HttpRequest):
...
+        return redirect(redirect_url)
+
+    @classmethod
+    def on_error(
+        cls,
+        project: Project,
+        request: HttpRequest,
+        err: str,
+        redirect_url: Optional[str] = None,
+    ):
```
mikkonie: Very minor thing, but the type hinting for both `on_success()` and `on_error()` is missing the `HttpResponse` return value.

projectroles/views.py
```diff
@@ -1672,14 +1672,60 @@ def _create_timeline_event(cls, project: Project, request: HttpRequest):
...
+            redirect to the home page)
+        """
+        p_type = get_display_name(project.type, title=False)
+        messages.error(request, f'Failed to delete {p_type}: {err}')
+        if not redirect_url:
+            redirect_url = reverse('home')
+        return redirect(redirect_url)
+
+    @classmethod
+    def get_redirect_url(cls, project: Project) -> str:
```
mikkonie: The `ProjectDeleteMixin` mixin has generic project deletion functionality which is called from both the UI views and the REST API views.

The `on_success()`, `on_error()` and `get_redirect_url()` methods contain purely UI specific functionality regarding to Django UI messages and UI view redirecting.

Hence, I believe these should instead be implemented in `ProjectDeleteView` and we should keep `ProjectDeleteMixin` for 100% UI/API agnostic operations.

projectroles/views.py
```diff
@@ -2012,24 +2068,55 @@ def post(self, *args, **kwargs):
...
-            redirect_url = self.get_redirect_url(project)
+                self.handle_delete(object, self.request)
+            return self.on_success(object, self.request)
         except Exception as ex:
             if settings.DEBUG:
                 raise ex
-            p_type = get_display_name(project.type, title=False)
-            messages.error(self.request, f'Failed to delete {p_type}: {ex}')
-        return redirect(redirect_url)
+            return self.on_error(object, self.request, str(ex))
```
mikkonie: If we expect the implementers of this base class to provide `on_success()` and `on_error()` implementations, this should be clearly documented.

I recommend adding some default implementations of those methods in this base class and documenting the need to implement one's own functionality in the docstring of each method.

projectroles/views.py
```diff
@@ -1970,38 +2016,48 @@ def post(self, request, **kwargs):
...
+    def get_type_display_name(self) -> str:
+        """Define a display name for the type of object"""
+        return 'object'
+
+    def get_context_data(self, *args, **kwargs):
+        """Override get_context_data()"""
+        context = super().get_context_data(*args, **kwargs)
+        context['warning_alert_message'] = self.get_warning_alert_message()
+        context['danger_alert_message'] = self.get_danger_alert_message()
+        context['type_display_name'] = self.get_type_display_name()
```
mikkonie: We could probably use `self.object.__class__.__name__` here, at least as the default if no override is provided?

projectroles/views.py
```diff
@@ -1672,14 +1672,60 @@ def _create_timeline_event(cls, project: Project, request: HttpRequest):
...
 
     @classmethod
-    def get_redirect_url(cls, project: Project) -> str:
+    def on_success(cls, project: Project, request: HttpRequest):
+        """
+        Called when project deletion is successful. It should print
+        a status message and redirect the user.
+
+        :param project: Project object
+        :param request: HttpRequest object
```
mikkonie: After moving the `on_success()` and `on_error()` methods into the new base view, we know we will aways have the request in `self.request`. At that point we can omit the `request` parameter from these methods.

projectroles/views.py
```diff
@@ -1970,38 +2016,48 @@ def post(self, request, **kwargs):
...
-    """Project deletion view"""
+    """Generic deletion view with confirmation form"""
 
-    model = Project
-    permission_required = 'projectroles.delete_project'
-    slug_field = 'sodar_uuid'
-    slug_url_kwarg = 'project'
+    template_name = 'projectroles/delete_confirmation_form.html'
+
+    def get_warning_alert_message(self) -> Optional[str]:
```
mikkonie: If we want to define the template here, we will need a bit more flexibility. E.g. in the SODAR use case, we may have multiple different warnings displayed at once depending on what the sample sheet data contains.

I see there are two options on how to handle this:

1. Allow the user to declare an arbitrary amount of messages in a list.
    * These would also supply the alert level to be rendered so it would have to be a list of dicts. E.g. `[{'level': 'danger', 'msg' '<p>Message which may also contain HTML</p>'}]`
    * Yes, note that the alerts also contain HTML :)
2. Allow the implementer of this view to declare their own template.
    * This will either be embedded into this view's deletion template, or override this template.
       + The latter approach would require us to provide the form part as an include template, which is of course trivial
    * This way they can setup and declare any alerts or other content they want without this view having to worry about it.
    * This would also make it much nicer to include HTML without providing ugly strings in Python
    
I'd prefer approach no.2, but let me know how you feel about it. I'd also be happy to brainstorm this or other issues in this review with you IRL.

projectroles/views.py
```diff
@@ -1970,38 +2016,48 @@ def post(self, request, **kwargs):
...
     DeleteView,
 ):
-    """Project deletion view"""
+    """Generic deletion view with confirmation form"""
 
-    model = Project
-    permission_required = 'projectroles.delete_project'
-    slug_field = 'sodar_uuid'
-    slug_url_kwarg = 'project'
+    template_name = 'projectroles/delete_confirmation_form.html'
```
mikkonie: If we decide on keeping this template (see below), I'd like to rename it into something like `confirm_delete_host.html` to better conform with typical Django template naming.

mikkonie (changes requested):
Good work, this works as intended. I have some suggestions in how to improve the distribution of responsibilities, how to make the template customization more flexible and the usual nitpicking ;)

I'd be happy to discuss any of these points with you in Teams/IRL, please ask if you have any questions or would like to suggest some other approaches.

projectroles/views.py
fmarotta: Done.

projectroles/views.py
fmarotta: Outdated: these methods don't exist anymore (see below).

projectroles/views.py
fmarotta: Completely agree. As we discussed IRL, I put the check for correct host name in the `dispatch()` method of the base view, and `ProjectDeleteMixin` is much cleaner.

After using `request.session.real_referrer` in the template, and implementing custom redirect logic in the `post()` method of `ProjectDeleteView`, there wasn't much need for `on_success()`, `on_error()`, and `get_redirect_url()`, so these methods have now been removed.

projectroles/views.py
fmarotta: Outdated.

projectroles/views.py
fmarotta: Good point, I've used that.

projectroles/views.py
fmarotta: Outdated

projectroles/views.py
fmarotta: As discussed IRL, I went for approach no.2. For custom warning messages, I created one default template  and one specialized for `ProjectDeleteView`, please let me know if this is what you meant.

projectroles/views.py
fmarotta: Done.

fmarotta:
Thanks for the review and the discussion IRL, I've addressed the comments. One nice thing is that the class implementing `HostConfirmDelete` doesn't have to implement anything: by default, the `post()` function of django's `DeleteView` is used, which simply calls `object.delete()`.

projectroles/templates/projectroles/confirm_delete_host.html
```diff
@@ -31,7 +25,7 @@ <h3>Confirm Deletion of {% get_display_name object.type True %}</h3>
            placeholder="Type the host name of this server (e.g. &quot;site.example.com&quot;) to confirm deletion." />
     <div class="btn-group pull-right">
       <a role="button" class="btn btn-secondary"
-          href="{% url 'projectroles:update' project=object.sodar_uuid %}">
+         href="{{ request.session.real_referer }}">
```
mikkonie, line 33: If we use `real_referer`, we need to add `HTTPRefererMixin` to `HostConfirmDeleteView` dependencies. That seems to be missing.

projectroles/views.py
```diff
@@ -1970,66 +1960,114 @@ def post(self, request, **kwargs):
...
     LoggedInPermissionMixin,
+    DeleteView,
+):
+    """Specialized deletion view with confirmation form asking for host name."""
+
+    #: Custom override for this view's template
+    template_name = 'projectroles/confirm_delete_host.html'
+
+    #: Custom template with warning messages shown in this view's form
+    custom_alerts_include = 'projectroles/_confirm_delete_host_alerts.html'
```
mikkonie: Since the included subtemplate can technically contain things other than alerts, we should name it less specifically. How about `template_include`?

projectroles/views.py
```diff
@@ -1970,66 +1960,114 @@ def post(self, request, **kwargs):
...
+
+    def get_type_display_name(self) -> str:
+        """Define a display name for the type of object"""
+        object = self.get_object()
+        return object.__class__.__name__
+
+    def get_context_data(self, *args, **kwargs):
+        """Override get_context_data()"""
+        context = super().get_context_data(*args, **kwargs)
+        context['custom_alerts_include'] = self.custom_alerts_include
```
mikkonie: Same comment as in the member variable, let's name this `template_include` instead to reflect the fact that the user can in fact add any HTML in that include.

docs/source/app_projectroles_api_django.rst
```diff
@@ -35,6 +35,10 @@ Base mixins and classes for building UI views can be found in
     :members:
     :show-inheritance:
 
+.. autoclass:: HostConfirmDeleteView
```
mikkonie: Minor cleanup thing: let's move this into the last one in the "Base UI View Classes" section. While I don't think there are written conventions for how to document these, it's nicer to first list the mixins and then the classes.

projectroles/templates/projectroles/confirm_delete_host.html
```diff
@@ -1,26 +1,20 @@
...
 </div>
 <div class="container-fluid sodar-page-container">
-  <div class="alert alert-danger" role="alert">
-    Are you sure you want to delete this {% get_display_name object.type %}?
-    {% if object.is_project %}
-       This will also delete all data within the project.
-    {% endif %}
-    This action can <strong>NOT</strong> be undone!
-  </div>
+  {% block custom_alerts %}
```
mikkonie: I was going to comment that we don't need the `{% block %}` directive here, as we provide the include template via a variable anyway.

..but that made me think: another way to do this would be to have that block here and let the user override the entire template with the existing `template` member variable.

That way the user could provide their own template, use `{% extends 'projectroles/confirm_delete_host.html' %}` and define their own content in `{% block custom_content %}`. This would remove the need to introduce a custom member variable and content data tweak. Also, it would make it more flexible to the user, as they could also override other parts of the template if needed, including even the entire form if they want to.

Let me know what you think? And feel free to experiment with `ProjectDeleteView` to see if this would actually work nicely.

projectroles/views.py
```diff
@@ -1970,66 +1960,114 @@ def post(self, request, **kwargs):
...
+    #: Custom override for this view's template
+    template_name = 'projectroles/confirm_delete_host.html'
+
+    #: Custom template with warning messages shown in this view's form
+    custom_alerts_include = 'projectroles/_confirm_delete_host_alerts.html'
+
+    #: URL for redirection after a successful deletion
+    success_url = reverse_lazy('home')
+
+    def get_type_display_name(self) -> str:
```
mikkonie: Maybe we should rename this and related variables into `object_display_name`? `type_display_name` seems a bit ambiguous.

mikkonie (dismissed):
This is looking good. I have one suggestion on how to possibly make this even a bit nicer, plus some very minor requests.

projectroles/templates/projectroles/confirm_delete_host.html
fmarotta: Right, I've added it

projectroles/views.py
fmarotta: Outdated by using the "extend" mechanism (see below)

projectroles/views.py
fmarotta: Outdated by using the "extend" mechanism (see below)

docs/source/app_projectroles_api_django.rst
fmarotta: Done

projectroles/templates/projectroles/confirm_delete_host.html
fmarotta: I agree, I think it makes sense. I originally added the `{% block %}` to give users the choice of using either "extend" or the custom include, but extending the template is more flexible anyway and probably more idiomatic in django. Besides, we'd have fewer templates to maintain.

projectroles/views.py
fmarotta: Sounds good.

mikkonie (approved):
No notes, this is looking real good now :)

Issue #1874, opened 2026-01-29
Batchupdateroles command cannot promote roles


fmarotta:
A test failure uncovered a typo in `batchupdateroles.py`: [the sign of the rank comparison](https://github.com/bihealth/sodar-core/blob/4ea5b476e28c6c0124728bb6881ca9d3b40ce974/projectroles/management/commands/batchupdateroles.py#L89C1-L97C19) seems to be inverted.

mikkonie:
Well spotted. If we do need to tag a v1.3.3 hotfix release before v1.4.0 is out, we should consider backporting this fix there. For now, it's fine to include it in the PR for #1869.

Issue #1869, opened 2026-01-21
Add transferroles management command


mikkonie:
We have recently discussed use cases where we want to transfer all roles from one user account to another, in cases where e.g. a user with multiple organizational accounts switches to using a different one, or they switch from using an LDAP account to OIDC, etc.

While owner roles can be transferred using `removeroles`, there is no command for transferring all project roles to a new user.

I'd like to have a management command which does the following (full spec below):

- Receive user names for `old_user` and `new_user`
- For each role assignment of `old_user`
    * Remove the assignment for `old_user`
    * Create a new role assignment of `new_user` of the same `role`
        + OR promote existing role assignment to a higher ranking `role`
- If optional flag is given, also deactivate `old_user` in the end

Eventually, we probably want to provide either a UI or some batch method for this, but for now a management command will suffice. Still, it's probably best to organize the code in a way where we can nicely refactor it into e.g. a view mixin later :)

We want to use existing mixins for role creation here instead of manipulating Django database objects directly. This is because the mixin helpers also handle e.g. timeline event creation and calling the project modify plugin API. This also requires a small change to one of the mixins, see spec below.

### Modify Mixins

- Add `notify` kwarg to `RoleAssignmentModifyMixin.modify_assignment()`
    * type=`boolean`, default=`True`
        + The default is very important!
    * Similar to kwarg in `RoleAssignmentDeleteMixin.delete_assignment()`
    * Add check for `notify` in `if request.user != user:`
        + We only want to create app alerts and send email if `notify==True`
        + **NOTE:** Timeline events should still be created
    * Remember to document kwarg in docstring
- Add `notify_new` kwarg to `RoleAssignmentOwnerTransferMixin.transfer_owner()`
    * We already have `notify_old` for notifying the old user
    * Add `notify_new` with similar logic to the aforementioned update
    * Default should be `True` here as well
- **NOTE:** No changes to tests should be required at this point, just make sure existing tests work

### Add Management Command

- Add management command `transferroles` in the `projectroles` app
    * You can use e.g. `removeroles` and `batchupdateroles` as examples
    * Use `ManagementCommandLogger` for logging, follow other conventions
    * Inherit relevant mixins
        + `RoleAssignmentModifyMixin`
        + `RoleAssignmentOwnerTransferMixin`
        + `RoleAssignmentDeleteMixin`
- Implement arguments
    * `-o, --old-user`: mandatory, corresponds to `SODARUser.username`
    * `-n, --new-user`: same as above
[...]

PR #1875, opened 2026-01-29
Add transferroles command (#1869)

fmarotta:
This PR adds a new management command, `transferroles`, which transfers the roles for all projects from one user to another. The implementation follows the spec in #1869. During this work, a separate problem in the `batchupdateroles` command was discovered (reported in #1874) and fixed as discussed IRL with @mikkonie.

I have also added some documentation, but it can be removed if the command should be internal (I was not sure about that).

### DIFF
--- a/docs/source/app_projectroles_usage.rst
+++ b/docs/source/app_projectroles_usage.rst
@@ -435,15 +435,16 @@ transferred to the parent category owner. Example:
 
     $ ./manage.py removeroles --user alice --owner bob
 
-Transfer All Roles from a User to Another
---------------------------
+Transfer Roles Between Users
+----------------------------
 
 The ``transferroles`` management command can be used to transfer roles for all
 projects from a user to another one (and optionally deactivate the old user).
 This is useful in cases where e.g. a user with multiple organizational accounts
 switches to using a different one, or they switch from using an authentication
 system to another. If the new user already had a lower-ranking role in a
-project, the new user is promoted to the higher role.
+project, the new user is promoted to the higher role. The old user can be
+deactivated by passing the `-d` flag.
 
 .. code-block:: console
 
--- a/docs/source/app_projectroles_usage.rst
+++ b/docs/source/app_projectroles_usage.rst
@@ -435,6 +435,20 @@ transferred to the parent category owner. Example:
 
     $ ./manage.py removeroles --user alice --owner bob
 
+Transfer All Roles from a User to Another
+--------------------------
+
+The ``transferroles`` management command can be used to transfer roles for all
+projects from a user to another one (and optionally deactivate the old user).
+This is useful in cases where e.g. a user with multiple organizational accounts
+switches to using a different one, or they switch from using an authentication
+system to another. If the new user already had a lower-ranking role in a
+project, the new user is promoted to the higher role.
+
+.. code-block:: console
+
+    $ ./manage.py transferroles --old-user alice --new-user bob
+
 
 User Status Checking
 --------------------
--- a/projectroles/management/commands/transferroles.py
+++ b/projectroles/management/commands/transferroles.py
@@ -1,0 +1,218 @@
+"""
+Transferroles management command for transferring all project roles
+to a new user.
+"""
+
+import sys
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from projectroles.management.logging import ManagementCommandLogger
+from projectroles.models import SODAR_CONSTANTS, Project, RoleAssignment
+from projectroles.views import (
+    RoleAssignmentDeleteMixin,
+    RoleAssignmentModifyMixin,
+    RoleAssignmentOwnerTransferMixin,
+)
+
+logger = ManagementCommandLogger(__name__)
+User = get_user_model()
+
+
+# SODAR constants
+PROJECT_ROLE_OWNER = SODAR_CONSTANTS['PROJECT_ROLE_OWNER']
+
+# Local constants
+USER_NOT_FOUND_MSG = 'User not found with username: {}'
+
+
+class Command(
+    RoleAssignmentOwnerTransferMixin,
+    RoleAssignmentModifyMixin,
+    RoleAssignmentDeleteMixin,
+    BaseCommand,
+):
+    help = 'Transfer all project roles to a new user.'
+
+    def _transfer_owner(
+        self,
+        project: Project,
+        new_user: User,
+        role_as: RoleAssignment,
+        p_title: str,
+    ) -> bool:
+        try:
+            with transaction.atomic():
+                self.transfer_owner(
+                    project=project,
+                    new_owner=new_user,
+                    old_owner_as=role_as,
+                    old_owner_role=None,
+                    notify_old=False,
+                    notify_new=False,
+                )
+                logger.info(
+                    f'Transferred ownership in {p_title} to {new_user.username}'
+                )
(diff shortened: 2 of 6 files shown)


docs/source/app_projectroles_usage.rst
```diff
@@ -435,6 +435,20 @@ transferred to the parent category owner. Example:
...
 
+Transfer All Roles from a User to Another
+--------------------------
+
+The ``transferroles`` management command can be used to transfer roles for all
+projects from a user to another one (and optionally deactivate the old user).
+This is useful in cases where e.g. a user with multiple organizational accounts
+switches to using a different one, or they switch from using an authentication
+system to another. If the new user already had a lower-ranking role in a
+project, the new user is promoted to the higher role.
```
mikkonie: Well documented! As a small improvement, we should also mention the optional `-d` argument for deactivating the old user.

docs/source/app_projectroles_usage.rst
```diff
@@ -435,6 +435,20 @@ transferred to the parent category owner. Example:
 
     $ ./manage.py removeroles --user alice --owner bob
 
+Transfer All Roles from a User to Another
```
mikkonie: I think this can be simplified as "Transfer Roles Between Users". Also, make sure the following line matches the title length, as many RST tools complain if they don't :)

projectroles/management/commands/transferroles.py
```diff
@@ -0,0 +1,221 @@
+"""
+Transferroles management command for transferring all project roles
+to a new user.
```
mikkonie: I think we should write "another user" here, as "new user" might imply that a new user needs to be created for this operation.

projectroles/management/commands/transferroles.py
```diff
@@ -0,0 +1,221 @@
...
+USER_NOT_FOUND_MSG = 'User not found with username: {}'
+
+
+class Command(
+    RoleAssignmentOwnerTransferMixin,
+    RoleAssignmentModifyMixin,
+    RoleAssignmentDeleteMixin,
+    BaseCommand,
+):
+    help = 'Transfer all project roles to a new user.'
```
mikkonie: Same here as in the module docstring, let's write "another user" to avoid confusion.

Note: in argument names and variables, `old_user` and `new_user` are still fine IMHO, since it's just a nice short way to denote the direction of transfer :)

projectroles/management/commands/transferroles.py
```diff
@@ -0,0 +1,221 @@
...
+        self,
+        project: Project,
+        new_user: User,
+        role_as: RoleAssignment,
+        p_title: str,
+    ) -> bool:
+        try:
+            with transaction.atomic():
+                self.delete_assignment(role_as, None, False)
+                self._create_role_or_promote(
```
mikkonie: Very minor comment, but I recommend moving `_create_role_or_promote()` before `_transfer_assignment()`. Although not required by Python, I prefer having functions defined before they are referred to in other functions where possible. This makes the class and/or module better readable and improves navigation.

projectroles/management/commands/transferroles.py
```diff
@@ -0,0 +1,221 @@
...
+                notify=False,
+            )
+            logger.info(
+                f'Created role "{role_as.role}" in {p_title} for '
+                f'{new_user.username}'
+            )
+        else:
+            # Promote user to higher role
+            assert new_user_roles.count() == 1
+            instance = new_user_roles.first()
```
mikkonie: In django, `instance` is usually used to newly created/updated objects while we are still in the creation or update process. Maybe we don't want to use this notation here, because it can be misleading.

As for what it should be, maybe just `new_user_role`?

As a sidenote: generally I'm all for keeping internal variable names short and concise, so you'll see me using just single or double word variable names a lot. However, as we handle multiple roles here, in this case it's better to be descriptive to make things maintainable :)

projectroles/tests/test_commands.py
```diff
@@ -1495,6 +1495,309 @@ def test_sync_ldap_existing(self):
...
+    RoleMixin,
+    RoleAssignmentMixin,
+    RemoteSiteMixin,
+    RemoteProjectMixin,
+    TestCase,
+):
+    """Tests for transferroles command"""
+
+    def setUp(self):
+        self.cmd_name = "transferroles"
```
mikkonie: Are these double quotes I see?! ;)

(I wish I could tell black to enforce my choice of quotes, maybe one day I will fork it..)

projectroles/tests/test_commands.py
```diff
@@ -1495,6 +1495,309 @@ def test_sync_ldap_existing(self):
...
+        self.assertEqual(
+            TimelineEvent.objects.filter(event_name='role_delete').count(), 1
+        )
+        self.assertEqual(
+            TimelineEvent.objects.filter(event_name='role_update').count(), 1
+        )
+        self.assertEqual(AppAlert.objects.count(), 0)
+        self.assertEqual(len(mail.outbox), 0)
+
+    def test_dont_demote_non_owner(self):
```
mikkonie: We can name this as just `test_demote_non_owner()`, as the expected failure is already defined in the nicely written docstring.

projectroles/tests/test_commands.py
```diff
@@ -1495,6 +1495,309 @@ def test_sync_ldap_existing(self):
...
+            self.category.get_owner().user, self.user_contributor_cat
+        )
+        self.assertEqual(
+            RoleAssignment.objects.filter(
+                user=self.user_contributor_cat
+            ).count(),
+            2,
+        )
+
+    def test_dont_demote_owner(self):
```
mikkonie: Same comment as in the earlier demote test, this can simply be `test_demote_owner()`.

mikkonie (dismissed):
Excellent work. I have a few comments, but these are all very minor cleanup stuff.

docs/source/app_projectroles_usage.rst
fmarotta: Done

docs/source/app_projectroles_usage.rst
fmarotta: Good call

projectroles/management/commands/transferroles.py
fmarotta: :+1:

projectroles/management/commands/transferroles.py
fmarotta: Done :)

projectroles/management/commands/transferroles.py
fmarotta: Alright, good to know

projectroles/management/commands/transferroles.py
fmarotta: Done

projectroles/tests/test_commands.py
fmarotta: 😬 they are gone! (Half of each, at least.)

This is something that [ruff](https://docs.astral.sh/ruff/settings/#format_quote-style) might support, we could look into it at some point (it's supposed to be a drop-in replacement). I'm happy to give it a try, if you want.

projectroles/tests/test_commands.py
fmarotta: Done

projectroles/tests/test_commands.py
fmarotta: And done.

mikkonie (approved):
Once again, good work! Merging.

PR #1879, opened 2026-02-02
Validate project/user kwargs in AppSettingsAPI.get()

### LINKED ISSUE #1873: AppSettingAPI.get() doesn't raise exception on missing project/user kwarg for PROJECT_USER setting
A pretty major bug I just noticed, which has apparently been in SODAR Core since forever. Oops!

For `PROJECT_USER` scope settings, we (obviously) require both `project` and `user` kwargs to be set.

In `AppSettingAPI.set()` this is correctly validated and an exception is raised if either is missing. However, `AppSettingAPI.get()` does not care about the presence of both kwargs. If not present, it happily returns the default value.

This is **not** by design: attempting to get a `PROJECT_USER` scope setting without both parameters should raise an exception regardless of whether the user has something other than the default selected.

It should be fixable by simply adding the following to the beginning of `get()` (as seen in `set()`):

```
s_def = cls.get_definition(name=setting_name, plugin_name=plugin_name)
cls._validate_project_and_user(s_def.scope, project, user)
```

I think we don't need a v1.3.3 hotfix release for this, having this in v1.4.0 is enough.

## Spec

- Add validation in `get()` (see above)
- Update `get()` docstring
    * Add: `:raise: ValueError if neither project nor user are set`
- Update tests
    * Ensure any combination of missing `project` or `update` kwargs with `get()` raises `ValueError`

## Tasks

- [ ] Add validation
- [ ] Update docstring
- [ ] Update tests

fmarotta:
This PR aims to fix #1873 by adding validation for `user` and `project` arguments of `AppSettingsAPI.get()`. Thanks @mikkonie for the support with troubleshooting the ensuing problems in `ProjectCreateView`.

The spec calls for raising a `KeyError` when the setting name doesn't exist; however, the `get_definition()` function, which is now called as part of the validation in `get()`, was raising a `ValueError`. I changed the error type to a `KeyError` because I think this was the intended behavior[^1], but please let me know otherwise.

The new behavior is as follows for both `get()` and `set()` :
- Setting name doesn't exist => `KeyError`
- Plugin name is wrong => `ValueError`
- App name is not set => `ValueError`
- Invalid user/project args based on scope => `ValueError`

**Warning**: I added the new commits to the branch `1873-validate-appsettings-get-fix` instead of creating a new one.

[^1]: The docstring of the `set()` function, which also calls `get_definition()`, says that `KeyError` is raised when the setting name doesn't exist, but actually a `ValueError` was raised, before this change. Tests have been updated accordingly.

### DIFF
--- a/projectroles/app_settings.py
+++ b/projectroles/app_settings.py
@@ -468,8 +468,8 @@ class AppSettingAPI:
         :param post_safe: Whether a POST safe value should be returned (bool)
         :param validate: Validate project and user args (bool, default=True)
         :return: Value
-        :raise: ValueError if nothing is found with setting_name, or if neither
-                project nor user are set
+        :raise: KeyError if nothing is found with setting_name
+        :raise: ValueError if neither project nor user are set
         """
         if validate:
             s_def = cls.get_definition(
@@ -599,7 +599,7 @@ class AppSettingAPI:
         :raise: ValueError if validating and value is not accepted for setting
                 type
         :raise: ValueError if neither project nor user are set
-        :raise: ValueError if setting name is not found in plugin specification
+        :raise: KeyError if setting name is not found in plugin specification
         """
         s_def = cls.get_definition(name=setting_name, plugin_name=plugin_name)
         cls._validate_project_and_user(s_def.scope, project, user)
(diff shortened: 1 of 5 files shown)


mikkonie (dismissed):
Good work on this one! However, it seems I was a bit confused about my spec. We should actually return `ValueError` for each of these error cases.

I guess I was thinking about the old implementation, where app settings were provided as a dict instead of a list of objects. In that case, `KeyError` would be in line with Python conventions.

If you can change the expected `KeyError` to `ValueError`, I'm happy to merge. Sorry about this! You were right to question me, I was mistaken :)

fmarotta:
Well, my assumption was incorrect :) I've changed everything back to `ValueError`, thanks for the review!

mikkonie (approved):
Looking good, will merge!

Issue #1848, opened 2025-12-09
Add client-side retrieval of siteinfo statistics


mikkonie:
On a large SODAR Core site with a lot of apps and data, loading the siteinfo view can be slow due to waiting for statistics queries server side.

While optimizing queries can help a bit, the best way to improve this would be to retrieve the app stats via an Ajax view in JQuery and update the DOM dynamically as results are returned.

In a nutshell, what I want to achieve is this:

- Stop querying for app plugin statistics in the server side view
- Add Ajax view from which to query the stats instead
- Call the Ajax view in JQuery, update DOM with results once received
    * One call to an Ajax view to retrieve all apps' statistics is good enough for now

### Add Ajax View

- Add Ajax view `PluginStatisticsAjaxView` in `siteinfo` app
    * Create new module `views_ajax`, see other apps for examples
    * Base view on `SODARBasePermissionAjaxView`
        + No project context needed as this is a site app
    * Use existing `siteinfo.view_info` permission for perm check
        + Only superusers should be able to access this
    * Define URL with URL name `ajax_stats` and route `stats`
    * Supported methods = `GET`, so implement functionality in `get()`
- Add view functionality
    * Retrieve all plugins
        * No need to separate these by project/site/backend like in the server side view, that's for the plugin list tab
        * Sort alphabetically by `plugin.title`
    * Retrieve stats for all plugins
        + **Hint:** Refactor functionality from `SiteInfoView._get_plugin_stats()`
        + Also make sure to return plugin title and icon to be displayed in DOM
    * If a plugin doesn't return any stats, no need to add anything to return data
    * Make sure to handle special cases without unhandled exceptions
        + Plugin not found with name
        + Exception encountered in plugin stats retrieval (see existing implementation)
- Add permission tests
    * These go in `test_permissions`
    * See existing site app Ajax view tests for examples, e.g. `adminalerts.tests.test_permissions`
- Add view tests
    * Retrieve statistics from some existing plugin which returns meaningful data
    * Retrieve statistics from a plugin which doesn't
    * Attempt retrieval with invalid plugin name
    * **TBD:** More?

### Update SiteInfoView and template

- Remove `_app_stats.html` includes from `site_info.html` template
- Remove `_get_plugin_stats()` and calls to it from view    
    * **NOTE:** We still need the plugins themselves in the context data to display their info, just remove the server-side statistics gathering
- (This might be a good point to update existing siteinfo view and UI tests to not fail :))
[...]

PR #1880, opened 2026-02-03
Add client-side retrieval of siteinfo statistics

fmarotta:
This PR attempts to address #1848 by implementing the spec described therein. I also had a couple of general questions:
- What's the best way to deal with template tags in JavaScript? Re-implement from scratch?
- Is it possible to dynamically disable some plugins or stats for testing purposes?


fmarotta:
Adding another one to the list...
- How to debug CI failures which don't occur locally? 😅

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,69 @@
+$(document).ready(function () {
+  $.get('ajax/stats').done(function (res) {
+    let cards = [];
```
mikkonie: Minor formatting note: the convention for JQuery in SODAR Core and SODAR is not to use semicolons at EOL, so we should remove these.

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,69 @@
...
+      if ('stats' in plugin) {
+        for (const [_, stat] of Object.entries(plugin.stats)) {
+          let info_link = '';
+          let stat_url = '';
+          if ('info_link' in stat) {
+            info_link = stat.info_link;
+          }
+          if ('url' in stat) {
+            stat_url = `
+              <dd class="col-md-9">
```
mikkonie: Instead of providing HTML as strings, I would recommend using JQuery objects whereever it is feasible. E.g. in this case doing `element.append($('<dd>').attr('class', 'classes-here'))` etc.

This might not be the best method for all cases and we can have exceptions where it doesn't make sense. In general I'd recommend this instead of adding in HTML as pure strings. It might be a bit tedious to write initially, but I've found this to make the code more manageable, as well as making it harder to accidentally insert invalid HTML.

Further along in this file I see some quite complex bits of HTML. If you feel it's less of a hassle to input those as strings instead of constructing a whole object hierarchy in JQuery, let me know and I'll trust your decision. But please do review which approach works better.

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,69 @@
...
+      let stats = [];
+      if ('stats' in plugin) {
+        for (const [_, stat] of Object.entries(plugin.stats)) {
+          let info_link = '';
+          let stat_url = '';
+          if ('info_link' in stat) {
+            info_link = stat.info_link;
+          }
+          if ('url' in stat) {
+            stat_url = `
```
mikkonie: The variable name `stat_url` is a bit misleading here, since it contains the full list element and its content, which may or may not contain a URL.

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,69 @@
+$(document).ready(function () {
+  $.get('ajax/stats').done(function (res) {
+    let cards = [];
+    for (const [plugin_id, plugin] of Object.entries(res)) {
+      let stats = [];
+      if ('stats' in plugin) {
+        for (const [_, stat] of Object.entries(plugin.stats)) {
+          let info_link = '';
```
mikkonie: In Javascript files, let's name variables in camelCase, as is standard.

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,69 @@
...
+          `);
+        }
+      } else if ('error' in plugin) {
+        stats.push(`
+          <div class="text-danger">
+            Unable to retrieve app statistics: ${ plugin.error }
+          </div>
+        `);
+      } else {
+        console.error('Unexpected plugin format from ajax/stats view.')
```
mikkonie: Maybe clarify this into "Unexpected response format for plugin {pluginName}".

Also, it might be a good idea to do a `console.dir()` on the response object after this error.

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,69 @@
+$(document).ready(function () {
+  $.get('ajax/stats').done(function (res) {
+    let cards = [];
+    for (const [plugin_id, plugin] of Object.entries(res)) {
```
mikkonie: I recommend renaming the `plugin` variable as something like `pluginData` or `pluginRes` to clarify that we are handling data related to a plugin, not e.g. a definition of the plugin.

Also, since `plugin_id` specifically refers to `plugin.name`, let's call the variable `pluginName` to avoid possible confusion.

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,69 @@
...
+        stats.push(`
+          <div class="text-danger">
+            Unable to retrieve app statistics: ${ plugin.error }
+          </div>
+        `);
+      } else {
+        console.error('Unexpected plugin format from ajax/stats view.')
+      }
+      cards.push(`
+        <div class="card" id="sodar-si-${ plugin_id }-app-stats-card">
```
mikkonie: This is from my original implementation, so it's my bad, but maybe this is a good time to improve this.

Using the plugin name as a part of the HTML ID attribute means we'll be adding underscores in the ID, which is not recommended.

Instead, we should add `sodar-si-app-stats-card` into the `class` attribute and introduce a separate `data-plugin-name` attribute containing the plugin name.

This requires refactoring tests, of course.

siteinfo/templates/siteinfo/site_info.html
```diff
@@ -182,7 +177,7 @@ <h4><i class="iconify" data-icon="mdi:sitemap"></i> Site Apps</h4>
         </div>
       </div>
       {# Enabled Backend Apps #}
-      <div class="card" id="sodar-si-site-apps-card">
+      <div class="card" id="sodar-si-backend-apps-card">
```
mikkonie, line 180: Good catch! :)

siteinfo/tests/test_ui.py
```diff
@@ -0,0 +1,156 @@
...
+app_settings = AppSettingAPI()
+
+
+class TestSiteInfoUI(UITestBase):
+    """Tests for the siteinfo view UI"""
+
+    def setUp(self):
+        super().setUp()
+        self.url = reverse('siteinfo:info')
+        self.serverside_stats_elements = [
```
mikkonie: We can shorten these variable names by removing the `side` bit, it's probably a bit redundant?

siteinfo/tests/test_ui.py
```diff
@@ -0,0 +1,156 @@
...
+        self.settings_elements = [
+            'sodar-si-settings-card-core',
+            'sodar-si-core-settings-card-app-bgjobs',
+            'sodar-si-core-settings-card-app-filesfolders',
+            'sodar-si-core-settings-card-app-timeline',
+            'sodar-si-core-settings-card-app-adminalerts',
+            'sodar-si-core-settings-card-app-tokens',
+        ]
+
+    def test_serverside_cards(self):
```
mikkonie: Same here, the `side` is not really required in naming (and if we really want to use it, we should write `server_side` but now I'm just getting pedantic :))

siteinfo/tests/test_ui.py
```diff
@@ -0,0 +1,156 @@
...
+            self.assertEqual(len(dt_children), len(dd_children))
+            self.assertGreater(len(dt_children), 0)
+
+    def test_clientside_cards_error(self):
+        """Test that client-side card errors are shown"""
+
+        def get_statistics_error(self):
+            raise ValueError('Invalid Stats')
+
+        # Monkey-patch the get_statistics() function to trigger an error
```
mikkonie, line 99: No notes, just complimenting on a good idea here :)

siteinfo/views.py
```diff
@@ -132,18 +132,15 @@ def _get_settings(cls, keys: list[str]) -> dict:
         return ret
 
     @classmethod
-    def _get_plugin_stats(cls, p_list: list) -> dict:
+    def _get_plugin_settings(cls, p_list: list) -> dict:
```
mikkonie, line 135: Again, no notes but a good call! I wrote this wrong in my spec, because of course we still need this to get the `info_settings`.

siteinfo/views_ajax.py
```diff
@@ -0,0 +1,61 @@
...
+from projectroles.views_ajax import SODARBasePermissionAjaxView
+from siteinfo.templatetags.siteinfo_tags import get_info_cls, get_info_val
+
+
+logger = logging.getLogger(__name__)
+plugin_api = PluginAPI()
+
+
+class PluginStatisticsAjaxView(SODARBasePermissionAjaxView):
+    """Site info ajax view"""
```
mikkonie: Let's describe the functionality a bit more, like "Plugin statistics retrieval ajax view"

mikkonie (dismissed):
Once again, good work. Works exactly as expected, I didn't find any functional problems. I have some comments, but they are mostly minor formatting things.

If you disagree with any of my suggestions or have questions, let's discuss.

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: Done

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: Sounds good, at first I thought it was less readable but it does catch certain errors :+1:

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: I have renamed it to `statValue`

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: Done

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: :+1:

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: Agreed

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: Sounds good! Should I also modify the server-side elements?

siteinfo/tests/test_ui.py
fmarotta: Done

siteinfo/tests/test_ui.py
fmarotta: Thanks :)

siteinfo/views_ajax.py
fmarotta: Done!

siteinfo/static/siteinfo/js/siteinfo.js
mikkonie: Good point! I see the same thing is done in `_plugin_settings.html`, didn't realize that. Yes, it would be nice to also get these updated.

siteinfo/static/siteinfo/js/siteinfo.js
```diff
@@ -0,0 +1,73 @@
...
+        'class': 'card sodar-si-app-stats-card',
+        'data-plugin-name': pluginName,
+      })
+      let cardHeader = $('<div>').attr('class', 'card-header').append(
+        $('<h4>').append([
+          $('<i>').attr({
+            'class': 'iconify',
+            'data-icon': pluginData.icon,
+          }),
+          pluginData.title + ' Statistics',
```
mikkonie: There is one problem which either didn't exist in the previous reviewed version, or I somehow missed it.

The margin between the icon and title is different between the server side elements and the ones built here in JQuery:

Image

I believe the easiest fix for this is simply to add a whitespace before `pluginData.title` here :)

mikkonie (changes requested):
Good work addressing the changes I raised. I noticed one minor cosmetic issue, also please do rework the server side elements with plugin name strings in ID attributes. Otherwise I'm good to merge.

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: Done

siteinfo/static/siteinfo/js/siteinfo.js
fmarotta: Ah, I missed that. Fixed now :)

mikkonie (approved):
Looks great! Once again good work. Merging.

PR #1884, opened 2026-02-04
Some proposals for updating the docs

fmarotta:
This PR advances some suggestions for the documentation, based on my reading as a new user/developer. The docs are already excellent and comprehensive, so it's hard to improve them. I just wanted to go through them and report a few comments.

1. The first paragraph of the index should convey more details about what SODAR Core does, so that the reader can decide whether they should use it or not.
2. I would find a list of the SODAR Core apps and their brief description useful to read before the installation instructions, so that I know what I will get before it's too late.
3. Since projects and roles are such a core concept, I would add a Project entry in the glossary.
4. The "getting_started" page currently doesn't have many details besides the requirements, and I find that it overlaps a bit with "app_projectroles_integration". So I would maybe move all the content from "app_projetroles_integration" into "getting_started".

I must say that going through the docs like this has been useful for me, so I am satisfied even if none of the suggestions are accepted :)


docs/source/glossary.rst
```diff
@@ -47,6 +47,11 @@ Glossary
         mirrored on the currently active site. This allows linking to remote
         projects on other sites where the user would have access.
 
+    Project
```
mikkonie, line 55: This is a good idea to add. Maybe also add "Category" with an explanation on how it differs from a project?

docs/source/getting_started.rst
```diff
@@ -4,47 +4,80 @@
...
+    differ in several ways from how SODAR Core is set up. This method is
+    recommended only for experienced Django developers.
+
+.. note::
+
+    For any other issues regarding the cookiecutter-django setup, see the
+    cookiecutter-django documentation.
+
+
+Installation on an Existing Site
```
mikkonie: I like the idea of moving the basic site setup stuff to "getting started", that makes sense.

However, maybe we should consider still having this existing site installation in a separate doc, maybe placing it in the "development" section?

Converting an existing Django site into a SODAR Core compatible one is most likely a very rare use case: we did it exactly once for Digestiflow, which was originally a standalone project.

This is also quite a bit more complicated than setting up a SODAR Core site from scratch, so seeing this wall of text in a "getting started" document might freak out some potential users :)

Maybe move this section to a separate integration document under the Development section and provide a link here under this section for further information?

Let me know what you think.

mikkonie (changes requested):
Good work! I had just a couple of suggestions for changes.

docs/source/glossary.rst
fmarotta: Done

docs/source/getting_started.rst
fmarotta: Good points, I totally agree

fmarotta:
Thanks, the docs have been updated. Sorry I still changed some things which were not directly related to your comments, maybe it's easier to see those if you only look at the second commit. In any case, let me know what you think now.

docs/source/glossary.rst
```diff
@@ -25,6 +25,11 @@ Glossary
         Django API provided by a backend app, to be dynamically imported and
         used by other SODAR Core based Django apps.
 
+    Category
+        If a project is a container for data, a category is a container for
+        projects and other categories. Categories can be used to organize
+        projects hierarchically, but cannot contain data.
```
mikkonie: I would rephrase the last part of this sentence as "are generally not expected to contain data".

It's technically very much possible to create an app which is enabled in categories and handles data in some sort. Some use cases might exist in e.g. providing a collective or summary interface for data under a category. In general this is not advisable though :)

mikkonie (changes requested):
Looking good, I had just one small comment. After handling that I'll merge.

docs/source/glossary.rst
fmarotta: Done

mikkonie:
This PR still appears in the "changes requested" state, but it looks like you handled the final comment I have. So all is looking good, will merge :)

mikkonie (approved):
Approved! Thank you for the improvements, this is very useful.

Issue #1861, opened 2026-01-13
Replace markupfield, pagedown and mistune with martor


mikkonie:
The `django-pagedown` package has not received updates in a long time. There are also some issues related to it, such as #669 and #1269. Similarly, `django-markupfield` has not been updated since 2021. It works fine for now but that may not last forever.

[Martor](https://github.com/agusmakmun/django-markdown-editor) (or `django-markdown-editor`) seems to be the FotM for markdown form fields, widgets and rendering. It appears to be actively maintained, too. We should give it a shot to switch to the more up-to-date package.

Another advantage of switching is that we should be able to remove the `mistune` dependency and use martor to also render markdown fields in non-form views. This should get us rid of the inconsistent formatting issues in #1269.

One possible issue is that according to their `setup.py`, Python 3.13 is not officially supported. Most likely this won't matter. Similarly, Django support is officially listed until 5.1, but since 5.2 is the LTS branch of that, we should be fine.

Another challenge may lie in migrating existing `MarkupField` fields into `MartorField` fields. Maybe it works easily with no intermediate steps or manual migration code needed, maybe not?

Furthermore, this package seems to be a Swiss army knife with a lot of support for all sorts of wild stuff. We should start with the minimal features and add support for more in the future if needed. What we care about:

- Django database model fields for markdown fields
- Markdown previews in forms
- Standard editor features in forms
- Rendering fields in templates similar to the preview

At least initially, we don't need to support e.g. image uploads or spell checking. When in doubt, disable :)

If there are ideas of an alternative package which would be even better than martor, I'm all ears. [django-markdownx](https://github.com/neutronX/django-markdownx) exists, but it has fewer features, a right-side preview layout I'm personally not crazy of and it doesn't seem to be in such active development as of late.

## Spec

- Add `martor` dependency
    * Pin to latest minor version (`>=1.7.15, <1.8` at the time of writing)
- Enable `martor` app
- Add relevant Django settings
    * Let's start with only the necessary ones for the aforementioned functionality
    * Make settings in `base.py` modifiable by environment variables
    * Also add hardcoded versions in `test.py`
- Replace `MarkupField` fields with `MartorField` fields
    * These are found in the `projectroles` and `adminalerts` apps
    * Create migrations as needed
    * Existing data needs to be preserved by migrations, of course
- Replace form widgets with martor widgets
- Customize martor widget to display relevant controls and match layout
    * Only enable critical features (see above)
[...]

PR #1887, opened 2026-02-13
Use martor for Markdown

fmarotta:
This PR replaces markupfield, pagedown and mistune with martor as specified in #1861.

## Main usage changes

- In forms, `SODARPagedownWidget` becomes `SODARMartorWidget`
- In templates, use `{{ object.markdown_field | safe_markdown }}` instead of `{% render_markdown ... %}`.

## Migrations

Some existing migrations are still using markupfield, so I haven't removed that dependency yet. I added new migrations for `projectroles.Project` and `adminalerts.AdminAlert` which transition the fields from markupfield to martor without loss of data. Here is how it works, specifically for the `description` field of `AdminAlert.
- Markupfield fields store the raw markdown as a string in the `description` column of the db; this column is a normal postgres TEXT column.
- In addition, markupfield creates two extra columns in the database: `_description_rendered`, which stores the precomputed html; and `description_markup_type`, which in our case is simply the literal "markdown" string.
- Martor, on the other hand, uses only one column, `description`, to store the raw markdown, and does the rendering on the fly; this column in postgres is also a simple TEXT type.
- The `description` column which stores the raw markdown text is compatible between markupfield and martor. Markupfield then adds some python sugar on top of that (such as the .raw attribute) to decide whether to access the raw markdown or the precomputed HTML.

The default django migration from markupfield to martor simply removes the two extra columns used by markupfield and declares the `description` column to be a `MartorField`, but the postgres data type does not change. In my test cases the data was not lost, but we should anyway double check in a larger instance :)

Notes:
- Martor docs say that when converting a text field to a martor field the data is preserved. [link](https://django-markdown-editor.readthedocs.io/en/latest/usage/models.html#converting-from-textfield)
- [Some](https://stackoverflow.com/questions/61267336/django-migration-file-depends-on-removed-django-app) [people](https://stackoverflow.com/questions/54729508/i-cant-remove-a-package-in-django-app-since-it-is-stored-in-migrations-files) in the internet suggest that squashing the migrations should allow us to remove the dependency on the old package

## Tricky bits

Martor includes a custom CSS theme in the pages where the markdown editor is used. By default, it is bootstrap v5. This affects the layout of the whole page, since SODAR Core uses bootstrap v4. To avoid that, I set the `MARTOR_ALTERNATIVE_CSS_FILE_THEME` setting.

## Styling

Here is a side-by-side comparison of the martor widget:
image

...vs the pagedown widget:
image

I'm happy to tweak the style a bit more if you have suggestions (and if martor allows me to do so :))

---

Once the styling is settled, I need to remember to
* [ ] update the screenshots in the documentation


mikkonie:
Thanks for the nice and detailed description along with the pull request. I'll get to reviewing it in a moment.

> [Some](https://stackoverflow.com/questions/61267336/django-migration-file-depends-on-removed-django-app) [people](https://stackoverflow.com/questions/54729508/i-cant-remove-a-package-in-django-app-since-it-is-stored-in-migrations-files) in the internet suggest that squashing the migrations should allow us to remove the dependency on the old package.

Yes, this is probably how we have to do it.

1. Squash migrations in SODAR Core v1.4.0, leave old migrations in (as is custom)
2. Remove old non-squashed migrations in v1.5.0
3. Add breaking changes note in v1.5.0 that any servers must first be upgraded to v1.4 before v1.5.

I will add a separate issue for squashing the migrations, as that should be done at the end of v1.4 development.

> Martor includes a custom CSS theme in the pages where the markdown editor is used. By default, it is bootstrap v5. This affects the layout of the whole page, since SODAR Core uses bootstrap v4. To avoid that, I set the MARTOR_ALTERNATIVE_CSS_FILE_THEME setting.

Well done. Had I realized this causes issues, I would have postponed this issue to v1.5 where I intend to upgrade the repo for Bootstrap v5. Oh well. As long as there's a workaround, this is fine :)

As for the styling, I'll see if I have any suggestions. I guess our aim should be to make it appear as close to the final elements as possible, but the most important thing is that all relevant Markdown elements get rendered in general (as that was an issue with the package used earlier)..

config/settings/base.py
```diff
@@ -236,6 +235,28 @@
...
 MEDIA_URL = '/media/'
 
+# MARTOR CONFIGURATION FOR MARKDOWN
+# ------------------------------------------------------------------------------
+# NOTE: Setting CSRF_COOKIE_HTTPONLY to `False` is required for AJAX uploads
+# Choose your preferred theme: "bootstrap" or "semantic"
+MARTOR_THEME = env.str('MARTOR_THEME', 'bootstrap')
+MARTOR_ENABLE_LABEL = env.bool('MARTOR_ENABLE_LABEL', True)
+MARTOR_ENABLE_CONFIGS = {
+    'imgur': env.str('MARTOR_ENABLE_CONFIG_IMGUR', 'false'),
```
mikkonie, line 245: Just a question: does Martor really require to provide these as `env.str()` instead of `env.bool()`? The latter would seem to make more sense..

config/settings/base.py
```diff
@@ -236,6 +235,28 @@
...
+    'jquery': env.str('MARTOR_ENABLE_CONFIG_JQUERY', 'false'),
+    'living': env.str('MARTOR_ENABLE_CONFIG_LIVING', 'true'),
+    'spellcheck': env.str('MARTOR_ENABLE_CONFIG_SPELLCHECK', 'false'),
+    'hljs': env.str('MARTOR_ENABLE_CONFIG_HLJS', 'false'),
+}
+# By default, martor loads bootstrap v5. We don't want that, so we supply an
+# alternative CSS file.
+MARTOR_ALTERNATIVE_CSS_FILE_THEME = env.str(
+    'MARTOR_ALTERNATIVE_CSS_FILE_THEME',
+    'projectroles/css/projectroles.css',
```
mikkonie: Shouldn't this be `martor.css` instead? Or the Bootstrap v4 CSS we use? After all, the `projectroles.css` only contains overrides for specifically named elements.

At the time of writing I haven't yet looked into how the styling works, so maybe I'm just misunderstanding this..

projectroles/templatetags/projectroles_common_tags.py
```diff
@@ -471,6 +473,11 @@ def get_remote_icon(project: Project, request: HttpRequest) -> str:
 @register.simple_tag
 def render_markdown(raw_markdown: str) -> str:
     """Markdown field rendering helper"""
+    logger.warning(
+        'The `render_markdown` template tag is deprecated and will be removed '
+        'in a future version, please use the `safe_markdown` filter '
```
mikkonie: We can specify that we will remove this in v1.5. Also, please add a separate issue for removing the tag and place it in the v1.5 milestone :)

mikkonie (changes requested):
Very good work once again. I played with this in dev and it works like a charm. I only have a few minor comments.

Regarding layout, I noticed a few things we could improve:

- The editor text appears a bit too large compared to general text in the layout. It should be made smaller to more or less match the text size in other form widgets.
- We should ensure the editor font is the same as we get for monospace styles and the `<code>` tag in our general CSS theme.
- Conversely, the text in the preview seems too small. I'd prefer it to match the size of the text as rendered in the views where this content is actually viewed.
- Maybe we should consider making the preview element maximum height limited to X, with scrolling applied if the content exceeds that? This is not critical, so if it can't be done trivially, feel free to add a new ticket for it and we can tackle it later.

Let me know if you have any questions on these ones or if you'd like to me take a crack at tweaking some things.

config/settings/base.py
fmarotta: [Yes](https://django-markdown-editor.readthedocs.io/en/latest/settings.html#:~:text=All%20values%20must%20be%20strings%3A%20%27true%27%20or%20%27false%27%2C%20not%20Python%20booleans), strings are required, although I also think bools would make more sense.

fmarotta: If it's less confusing, we could maybe just use one environment variable, `MARTOR_ENABLE_CONFIGS`, which takes a json string, instead of having multiple variables, one for each config key.

config/settings/base.py
fmarotta: I agree, using `martor.css` or Bootstrap v4 would make sense. I had tried both, but faced some issues.

The `martor.css` is meant to be loaded directly in the widget using the `Media` class (see [here](https://github.com/bihealth/sodar-core/pull/1887/changes#diff-5b60ded14a7f0f987bceb505d1a1e0555d6464724fef4deb3bf95e9f9bee3f59L376). If we load it here, another CSS file (`martor.bootstrap.min.css`) will be loaded later and overwrite the CSS rules in `martor.css`. In other words, I think the order in which martor loads CSS is:
```
$MARTOR_ALTERNATIVE_CSS_FILE_THEME -> martor.bootstrap.min.css -> martor.css
```
however, if `$MARTOR_ALTERNATIVE_CSS_FILE_THEME == martor.css`, the file is cached and is not loaded the second time. If `martor.css` is not loaded last, the rules in `martor.bootstrap.min.css` take precedence. We could of course work around this by using `!important` or more specific selectors..

Loading Bootstrap v4 in my opinion would be a better solution, because it follows the spirit of the "alternative CSS theme" more closely. The reason why I didn't use that (and I may very well be mistaken) is because I didn't find a local Bootstrap CSS file in the development environment. I think the docker image for production may save the files in `static/local`, but the local site just uses the CDN. So, I wasn't sure whether I should download Bootstrap v4 locally and add it to the repository. In the end, I decided to load a "neutral" CSS file, the `projectroles.css`.

Another option could have been to set it to a non-existent path, but then users would probably see a "file not found" error in the browser console and that wouldn't be very nice.

projectroles/templatetags/projectroles_common_tags.py
fmarotta: Done

config/settings/base.py
fmarotta: Update: I've tried to point that variable to the URL of the CDN for the Bootstrap version we use. However, since martor loads this file after `projectroles.css`, all the custom rules in projectroles are ignored.

As per our IRL discussion, I've created a `martor_theme.css` file and pointed the variable to that path. Let me know what you think now :)

fmarotta:
Thanks for your review and the IRL talk. I've updated the styles and compared the editor preview widget with `sodar-markdow-content` elements, now AFAICT they look very similar, except for the background color. I've used a markdown demo from [here](https://markdown-it.github.io/) to check the rendering of all elements. If we need to support some of the more exotic markdown features we can enable them through `MARTOR_MARKDOWN_EXTENSIONS` in the settings.

---

Martor features which would make our life easier (maybe worth to consider as tickets for martor):
- Use booleans instead of "true"/"false" strings for config settings
- Use `rem` units for heading elements instead of px in `martor.bootstrap.min.css`
- Allow users to change the class of `martor-preview` elements
- Support different versions of Bootstrap

config/settings/base.py
mikkonie: As we discussed this IRL, let's do it as they have designed it. Providing dictionaries in e.g. Ansible is a bit of a hassle. A bit of a weird decision on the part of Martor devs, but it's not our fault :)

mikkonie (approved):
Very good work. There are a couple of very minor styling things I could still try tweaking myself, as they are a bit hard to describe and I'm not sure what the correct solution is. I'll take a look at them when cleaning up the v1.4.0 release.

As for this PR, consider it merged :)

PR #1888, opened 2026-02-13
Add user active status to remote sync 

### LINKED ISSUE #1882: Add user active status to remote sync
While thinking about some other remote sync issues, I realized that despite adding proper handling of user activity status in SODAR Core v1.1+, we don't sync the active status of users from source to target sites. We should add this.

This is a non-breaking change which will bump the `projectroles.sync` API to v2.1.

We need to update the activity for both the source server and target server functionality:

- Source server
    * Add active status to user info if called with API version 2.1 or above
    * Done in `_add_user()`, which is called by `get_source_data()`
- Target server
    * Update active status of user if present and changed 
    * Done in `_sync_user()`, which is called by `sync_remote_data()`
    * Failsafe: do not update activity for target site superusers
         + This might result in unwanted lockouts for admins :)

Side note: I apologize in advance for making someone to look into the terrible remote sync API, which was a quick two-afternoon hack that is somehow still in use. Some day we will rewrite the whole thing with proper serializers etc. Today is not that day.

## Spec

- Update API versions in `projectroles.views_api`
    * `SYNC_API_DEFAULT_VERSION = '2.1'`
    * Add `2.1` to `SYNC_API_ALLOWED_VERSIONS`
- Update `_add_user()`
    * For user data, add `User.active` as `active` field with boolean value
    * Only add field to user dict if request API version is `>=2.1`
- Update tests in `TestGetSourceData`
    * With default API version, `active` should be included for users
    * Add test with API v2.0 and ensure no `active` field is returned for user
- Update `_sync_user()`
    * Base functionality should work without requiring changes
        + If called with API version `>=2.1`, the `active` field is included
        + All fields (except `additional_emails`) get passed to `_update_object()`
    * Add check for existing user superuser status
        + If `user.is_superuser`, pop `active` field if present
- Update tests in `TestSyncRemoteDataCreate` or `TestSyncRemoteDataUpdate`
    * (It shouldn't really matter in which class the user updating is tested)
    * Ensure `active=False` status is updated for user
    * Ensure status is not updated for superuser
- Update documentation
    * Add note on non-superuser activity status updating in `app_projectroles_usage.rst` in the "As Target Site" section

## Tasks

- [ ] Update API versions
- [ ] Update `_add_user()`
- [ ] Update `get_source_data()` tests
- [ ] Update `_sync_user()`
- [ ] Update `sync_remote_data()` tets
- [ ] Document

fmarotta:
This PR attempts to fix #1882 by following the spec described therein.

There are two changes:
- The `is_active` key is removed from `updated_fields` instead of from `user_data`
- The function `_sync_user()` was split in two

The reason for the first is that I've seen that something similar is done for `additional_emails`, where the field is removed from the dictionary and then added at the end. I wasn't sure whether this is needed also for `is_active`, but in order to avoid having to re-add that field at the end, I opted to remove it from `updated_fields`.

The reason for splitting `_sync_user()` was that otherwise the CI would have failed with an error from flake8 about the cyclomatic complexity being too high (19).

The tests are passing but I haven't played with the feature interactively yet, so the PR is marked as draft for the time being.

### DIFF
--- a/docs/source/app_projectroles_usage.rst
+++ b/docs/source/app_projectroles_usage.rst
@@ -656,6 +656,10 @@ Alternatively, the following management command can be used:
     users must be manually created by a target site admin in order for their
     data and roles to be synchronized.
 
+.. note::
+
+    Activity status for all users except superusers is synchronized.
+
 Project Detail View Links
 -------------------------
 
--- a/projectroles/tests/test_remote_project_api.py
+++ b/projectroles/tests/test_remote_project_api.py
@@ -77,6 +77,7 @@ SOURCE_USER_FIRST_NAME = SOURCE_USER_NAME.split(' ')[0]
 SOURCE_USER_LAST_NAME = SOURCE_USER_NAME.split(' ')[1]
 SOURCE_USER_EMAIL = SOURCE_USER_USERNAME.split('@')[0] + '@example.com'
 SOURCE_USER_UUID = str(uuid.uuid4())
+SOURCE_USER_IS_ACTIVE = True
 
 SOURCE_USER2_DOMAIN = SOURCE_USER_DOMAIN
 SOURCE_USER2_USERNAME = 'source_user2@' + SOURCE_USER_DOMAIN
@@ -86,6 +87,7 @@ SOURCE_USER2_FIRST_NAME = SOURCE_USER2_NAME.split(' ')[0]
 SOURCE_USER2_LAST_NAME = SOURCE_USER2_NAME.split(' ')[1]
 SOURCE_USER2_EMAIL = SOURCE_USER2_USERNAME.split('@')[0] + '@example.com'
 SOURCE_USER2_UUID = str(uuid.uuid4())
+SOURCE_USER2_IS_ACTIVE = True
 
 SOURCE_USER3_DOMAIN = SOURCE_USER_DOMAIN
 SOURCE_USER3_USERNAME = 'source_user3@' + SOURCE_USER_DOMAIN
@@ -95,6 +97,7 @@ SOURCE_USER3_FIRST_NAME = SOURCE_USER3_NAME.split(' ')[0]
 SOURCE_USER3_LAST_NAME = SOURCE_USER3_NAME.split(' ')[1]
 SOURCE_USER3_EMAIL = SOURCE_USER3_USERNAME.split('@')[0] + '@example.com'
 SOURCE_USER3_UUID = str(uuid.uuid4())
+SOURCE_USER3_IS_ACTIVE = False
 
 SOURCE_USER4_DOMAIN = SOURCE_USER_DOMAIN
 SOURCE_USER4_USERNAME = 'source_user4@' + SOURCE_USER_DOMAIN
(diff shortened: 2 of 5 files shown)


projectroles/tests/test_remote_project_api.py
```diff
@@ -1086,6 +1140,10 @@ def test_create_assert_category_roles(self):
         new_user2 = User.objects.get(username=SOURCE_USER2_USERNAME)
         new_user3 = User.objects.get(username=SOURCE_USER3_USERNAME)
         new_user4 = User.objects.get(username=SOURCE_USER4_USERNAME)
+        self.assertTrue(new_user.is_active)
```
mikkonie, line 1138: Since you define the `SOURCE_USER*_IS_ACTIVE` constants, it would be consistent to use them here too?

Or, alternatively, since these are simple bool values, maybe get rid of the consts and simply refer to them with `True`/`False`? You can make the call :)

docs/source/app_projectroles_usage.rst
```diff
@@ -656,6 +656,10 @@ Alternatively, the following management command can be used:
     users must be manually created by a target site admin in order for their
     data and roles to be synchronized.
 
+.. note::
+
+    Activity status for all users except superusers is synchronized.
+
```
mikkonie, line 662: It seems I had forgotten one thing from my spec: we should also document the API changes in `major_changes.rst`. This should be done as follows:

-  Add "REST API View Changes" section under "Breaking Changes"
   * Yes, this is not actually breaking, but that's how I have organized things for some reason :)
- Document version number change and addition of `is_active` for users
    * See v1.2.0 release notes for an example on how to document such changes

Sorry about missing this in my spec!

projectroles/tests/test_remote_project_api.py
```diff
@@ -843,6 +892,7 @@ def setUp(self):
         self.admin_user = self.make_user(settings.PROJECTROLES_DEFAULT_ADMIN)
         self.admin_user.is_staff = True
         self.admin_user.is_superuser = True
+        self.admin_user.save()
         self.maxDiff = None
```
mikkonie: This `self.maxDiff = None` line seems to be a accidental remnant from some old debugging activity. It's not your doing, but feel free to remove it while cleaning up :)

mikkonie (changes requested):
Good work! I had a couple of minor comments, one of which is due to me forgetting to add something to the spec. Apologies for that.

projectroles/tests/test_remote_project_api.py
fmarotta: I'll get rid of the constants then :) my reasoning for not reusing them was to make the tests more robust, e.g. if 10 years from now we forget about this feature and change all the constants to `True`, the tests will pass but the case where users are inactive will never occur. So I wanted to make sure that at least one user would have `False` by hard-coding the value in the test[^1].

[^1]: ...only to find out 20 years from now that I even forget about this and change the `False` into `True`.

docs/source/app_projectroles_usage.rst
fmarotta: No problem, done!

projectroles/tests/test_remote_project_api.py
mikkonie: I agree that the robustness is a good point. As I wrote, it would also be very much acceptable to check against the constants here, instead of true/false. But it is indeed your call.

mikkonie (approved):
Looks good! I took the liberty to correct one tiny mistake in `major_changes.rst`, as it listed the v2.1 changes to the sync api as "breaking", as a .1 API bump is by definition non-breaking :)

Other than that, no notes, will merge.

Issue #1191, opened 2023-04-03
Add project limiting to search


mikkonie:
We should be able to limit search results by project UUID. Not sure why this was never implemented.

**Update 2026-02:** This has also recently been requested (again) by a SODAR power user. Let's go ahead and do it.

How this should work:

- User makes a search with keyword `project:{Project.sodar_uuid}`
   * Example: `search term project:6bffad6d-93ad-402b-bb7d-0da269f9ee5b`
- Search only returns items related to the project with the given UUID.
- We should also update search functionality in SODAR Core apps to handle this keyword.

By using the `keywords` kwarg in search, this will not be breaking. We should first do #1894 though to ensure the kwarg is presented correctly in the API.

This will not be breaking, as the `keywords` kwarg already exists. However, this hands over the responsibility for providing expected results to the apps implementing `search()`.

We *could* also add filtering in `ProjectSearchMixin._get_app_results()`. However, we can't guarantee that every model in apps implemented by SODAR Core based sites has a `ForeignKey` with the name `project`. Also, some apps require logic unknown to SODAR Core in filtering results, like searching for iRODS files in SODAR. Last but not least, this would perform much worse than apps handling the filtering themselves. Hence, IMHO, we should trust the apps to handle this and simply document the expected behaviour. Let me know if you disagree.

## Spec

**Update 2026-03-05:** Updated according to discussions related to #1899.

- Add `projects` argument to `ProjectAppPlugin.search()`
    * Type = `QuerySet[Project]`
    * Mandatory, goes after `user`
- Populate `projects` argument in `ProjectSearchResultsView`
    * IF `project` keyword is given, QuerySet is `Project.objects.filter(sodar_uuid=keyword_value)`
    * ELSE `Project.objects.all()`
    * (More fancy filtering will be done in #1898 and beyond..)
- Pass new arg to plugins in `ProjectSearchMixin`
- Update `find()` methods in object managers to filter projects down to the `projects` arg
    * `projectroles.models.ProjectManager`
    * `filesfolders.models.FilesfoldersManager`
    * `timeline.models.TimelineEventManager`
- Handle invalid UUIDs gracefully, e.g. don't return anything
- Update tests for all three apps to ensure filter is properly handled
    * Remember to also test with invalid/missing UUID :)
- Update documentation
    * `app_projectroles_usage.rst` under "Search"
        + Add mention of `project` keyword
        + Maybe reorganize this nicely together with the `type` keyword?
[...]

PR #1895, opened 2026-02-25
Fix type hint and docstring for keywords argument in ProjectAppPlugin.search()

### LINKED ISSUE #1894: Incorrect type hint and docstring description for keywords in ProjectAppPlugin.search()
The `ProjectAppPlugin.search()` method contains a `keywords` kwarg, which was defined in the API but has so far not been used.

I've been looking into #1191, where we would want to put `keywords` to use by providing a `project` keyword with the project UUID as value.

I noticed that both the type hint and docstring for this method are incorrect:

```
    def search(
        self,
        search_terms: list[str],
        user: User,
        search_type: Optional[str] = None,
        keywords: Optional[list[str]] = None,
    ) -> list['PluginSearchResult']:
        """
        Return app items based on one or more search terms, user, optional type
        and optional keywords.

        :param search_terms: Search terms to be joined with the OR operator
                             (list of strings)
        :param user: User object for user initiating the search
        :param search_type: String
        :param keywords: List (optional)
        :return: List of PluginSearchResult objects
        """
```

Keywords are expected to be provided as an optional `dict` of key/value pairs, not a list of strings. The type hint and docstring should be updated to reflect this.

This fix should also be copied to `filesfolders.plugins` and `timeline.plugins`, where we implement the method.

Curiously, the object manager `find()` methods these plugin calls have the type hint down correctly. I guess my brain short circuited when I was adding type hints to the repo. Can I lie and say I was vibe coding it to avoid responsibility? :)

**Note:** This should **not** be considered a breaking change, as the `keywords` is (as far as I recall?) not actually used anywhere yet.

## Tasks

- [ ] Fix `projectroles.plugins.ProjectAppPlugin.search()` type hint and docstring
- [ ] Apply changes to `filesfolders`
- [ ] Apply changes to `timeline`
- [ ] Update example in docs if there is one?

fmarotta:
This PR attempts to fix #1894. I have also taken the liberty to updat the type hint and docstring in `ProjectSearchMixin._get_app_results()`. The example in the docs was updated to reflect the key/value pairs nature of `keywords`, but I think more changes to the docs will be done for #1191 soon.

### DIFF
--- a/docs/source/dev_project_app.rst
+++ b/docs/source/dev_project_app.rst
@@ -637,7 +637,7 @@ See the signature of ``search()`` in
     - You can specify supported types in the plugin's ``search_types`` list.
     - Examples: ``file``, ``sample``..
 ``keywords``
-    - Special search keywords, e.g. "exact".
+    - Special search keywords as key/value pairs.
     - **NOTE:** Currently not implemented.
 
 .. note::
--- a/filesfolders/plugins.py
+++ b/filesfolders/plugins.py
@@ -162,7 +162,7 @@ class ProjectAppPlugin(ProjectAppPluginPoint):
         search_terms: list[str],
         user: User,
         search_type: Optional[str] = None,
-        keywords: Optional[list[str]] = None,
+        keywords: Optional[dict] = None,
     ) -> list[PluginSearchResult]:
         """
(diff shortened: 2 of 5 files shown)


mikkonie (approved):
No notes, good work :)

PR #1896, opened 2026-02-26
Add project limiting to search

### LINKED ISSUE #1191: Add project limiting to search

fmarotta:
This PR attempts to fix #1191 by implementing the spec therein. It has conflicts with #1895 but I can rebase it after that is merged.

### DIFF
--- a/docs/source/major_changes.rst
+++ b/docs/source/major_changes.rst
@@ -41,12 +41,15 @@ recommended to manipulate ``AppSetting`` objects directly.
 Plugin API search() Changes
 ---------------------------
 
-The ``search()`` function implemented by plugins already had the ``keywords``
-argument in its signature, but so far it was not used. This release implemented
-support for the ``project:uuid`` keyword. All SODAR Core apps are now expected
-to support this keyword as well, which, if present in the ``keywords``
-dictionary, should restrict the search to the project or category specified by
-the UUID.
+The signature of the ``search()`` function implemented by plugins has changed:
+it now takes an extra positional argument, ``projects``, which contains a
+``QuerySet`` of projects within which the search should be restricted. All
+SODAR Core apps are expected to honor this filtering criterion. Furthermore,
+the ``keywords`` dictionary (passed to ``search()`` as a kwarg) can contain
+a ``project:uuid`` key-value pair which identifies the user-provided project
+or category where the search is restricted. The projects are pre-fetched for
+efficiency and passed to ``search()`` through the ``projects`` argument, but
+apps can optionally also make use of the original UUID.
 
 Previously Deprecated Features Removed
 --------------------------------------
--- a/docs/source/major_changes.rst
+++ b/docs/source/major_changes.rst
@@ -41,10 +41,12 @@ recommended to manipulate ``AppSetting`` objects directly.
 Plugin API search() Changes
 ---------------------------
 
-This release implemented support for the ``project:uuid`` keyword in the plugin
-``search()`` function. All SODAR Core apps are now expected to support this
-keyword as well, which should restrict the search to the project or category
-specified by the UUID.
+The ``search()`` function implemented by plugins already had the ``keywords``
+argument in its signature, but so far it was not used. This release implemented
+support for the ``project:uuid`` keyword. All SODAR Core apps are now expected
+to support this keyword as well, which, if present in the ``keywords``
+dictionary, should restrict the search to the project or category specified by
+the UUID.
 
 Previously Deprecated Features Removed
 --------------------------------------
--- a/docs/source/major_changes.rst
+++ b/docs/source/major_changes.rst
@@ -38,6 +38,14 @@ access the ``user_modifiable`` variable of an app setting, you should instead
 use ``AppSettingAPI.get_definition()``. Note that it is not generally
 recommended to manipulate ``AppSetting`` objects directly.
 
+Plugin API search() Changes
+---------------------------
+
+This release implemented support for the ``project:uuid`` keyword in the plugin
+``search()`` function. All SODAR Core apps are now expected to support this
+keyword as well, which should restrict the search to the project or category
+specified by the UUID.
+
 Previously Deprecated Features Removed
 --------------------------------------
 
--- a/filesfolders/models.py
+++ b/filesfolders/models.py
@@ -5,6 +5,7 @@ import uuid
 from typing import Optional
 
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models import Q, QuerySet
 
@@ -74,6 +75,16 @@ class FilesfoldersManager(models.Manager):
         :return: QuerySet of BaseFilesfolderClass objects
         """
         objects = super().get_queryset().order_by('name')
+        if keywords and 'project' in keywords:
+            try:
+                project = Project.objects.get(sodar_uuid=keywords['project'])
+                objects = objects.filter(
+                    project__full_title__startswith=project.full_title
+                )
+            except Project.DoesNotExist:
+                return objects.none()
+            except ValidationError:
+                return objects.none()
         term_query = Q()
         for t in search_terms:
             term_query.add(Q(name__icontains=t), Q.OR)
--- a/filesfolders/models.py
+++ b/filesfolders/models.py
@@ -80,7 +80,7 @@ class FilesfoldersManager(models.Manager):
         objects = (
             super()
             .get_queryset()
-            .filter(Q(project__in=projects) | Q(project__isnull=True))
+            .filter(Q(project__in=projects))
             .order_by('name')
         )
         term_query = Q()
--- a/filesfolders/models.py
+++ b/filesfolders/models.py
@@ -5,7 +5,6 @@ import uuid
 from typing import Optional
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models import Q, QuerySet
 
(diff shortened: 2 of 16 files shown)


filesfolders/tests/test_models.py
```diff
@@ -150,28 +169,95 @@ def test_initialization(self):
             'sodar_uuid': self.folder.sodar_uuid,
         }
         self.assertEqual(model_to_dict(self.folder), expected)
+        expected = {
```
mikkonie: Since this test case is specifically there to test model initialization and `other_folder` is there to test `find()`, we don't necessarily have to test its initialization here.

filesfolders/tests/test_models.py
```diff
@@ -137,6 +144,18 @@ def setUp(self):
             owner=self.user_owner,
             description='description',
         )
+        # Make other project
+        self.other_project = self.make_project(
```
mikkonie: Minor note, but if it's a second object for general purposes, we usually name these e.g. `self.project2`. More compact and faster to edit :)

(If there is a case where we need to set up several objects for specific purposes, explicitly naming them is of course recommended)

filesfolders/tests/test_models.py
```diff
@@ -150,28 +169,95 @@ def test_initialization(self):
...
 
     def test_find_fail(self):
         """Test FilesfoldersManager find() with a non-existing Folder"""
-        objects = Folder.objects.find(['Jaix1azu'])
+        objects = Folder.objects.find(['Jaix1azu'], Project.objects.all())
+        self.assertEqual(len(objects), 0)
+
+    def test_find_with_keywords(self):
+        """Test FilesfoldersManager find() with project keyword"""
+        # Should work
```
mikkonie: I would split this into four separate unit tests. The more atomic we can get with unit tests the better and these seem to be quite distinct cases here :)

filesfolders/models.py
```diff
@@ -64,16 +64,25 @@ class FilesfoldersManager(models.Manager):
...
         :param search_terms: Search terms (list of strings)
+        :param projects: QuerySet of projects where the terms are searched
         :param keywords: Optional search keywords as key/value pairs (dict)
         :return: QuerySet of BaseFilesfolderClass objects
         """
-        objects = super().get_queryset().order_by('name')
+        objects = (
+            super()
+            .get_queryset()
+            .filter(Q(project__in=projects) | Q(project__isnull=True))
```
mikkonie: The `project` field is required for `BaseFilesfoldersClass`, so `project__isnull` should probably not be needed here. Or is there some case which necessiates this?

projectroles/tests/test_models.py
```diff
@@ -1098,68 +1098,189 @@ def setUp(self):
...
         """Test find() with multiple terms for different fields"""
-        result = Project.objects.find(['project', 'xxx'])
+        result = Project.objects.find(
+            ['project', 'xxx'], projects=self.all_projects_qs
+        )
+        self.assertEqual(len(result), 3)
+
+    def test_find_with_project_filtering(self):
+        """Test find() restricted to a category"""
+        result = Project.objects.find(
```
mikkonie, line 1220: This could also be split to multiple atomic tests similarly to the one in filesfolders.

projectroles/tests/test_models.py
```diff
@@ -1098,68 +1098,189 @@ def setUp(self):
...
+        """Test find() with project_type and project filtering"""
+        result = Project.objects.find(
+            ['Test'],
+            projects=self.category_sub_qs,
+            project_type=PROJECT_TYPE_CATEGORY,
+            keywords={'project': self.category_sub.sodar_uuid},
+        )
+        self.assertEqual(list(result), [self.category_sub])
+
+        result = Project.objects.find(
```
mikkonie, line 1261: Same here, this could be two separate test cases.

projectroles/tests/test_views.py
```diff
@@ -492,7 +549,82 @@ def test_post_advanced(self):
...
+        )
+        self.assertEqual(
+            response.context['search_keywords'],
+            {'project': str(project_new.sodar_uuid)},
+        )
+        self.assertEqual(response.context['search_type'], None)
+        # Only the project matching 'xxx' should be returned
+        self.assertEqual(len(response.context['project_results']), 1)
+        # Get rid of the temporary project
+        project_new.delete()
```
mikkonie: Manual deletion is not needed, the database is reset between tests.

projectroles/tests/test_views.py
```diff
@@ -588,6 +756,74 @@ def test_get_omit_app(self):
             self.assertEqual(response.status_code, 200)
             self.assertEqual(len(response.context['app_results']), 1)
 
+    def test_get_app_search_results_with_keywords(self):
```
mikkonie: I would also recommend splitting this test. It seems to convert into three different test quite nicely.

projectroles/models.py
```diff
@@ -98,13 +99,16 @@ def find(
...
 
         :param search_terms: Search terms (list)
+        :param projects: QuerySet of projects where the terms are searched
         :param keywords: Optional search keywords as key/value pairs (dict)
         :param project_type: Project type or None
         :return: QuerySet of Project objects
         """
-        projects = super().get_queryset().order_by('title')
+        objects = (
+            super().get_queryset().filter(pk__in=projects).order_by('title')
```
mikkonie: It seems the `order_by()` here is redundant, since we order the objects again when returning them. This is my doing from some older implementation, but let's get it fixed while we're cleaning up this PR :)

timeline/tests/test_models.py
```diff
@@ -341,52 +341,128 @@ def test_add_object_no_name(self):
...
         """Test TimelineEvent.find() with user full name not found"""
         self.user_owner.name = 'Owner User'
         self.user_owner.save()
-        objects = TimelineEvent.objects.find(['Other User'])
+        objects = TimelineEvent.objects.find(
+            ['Other User'], Project.objects.all()
+        )
+        self.assertEqual(len(objects), 0)
+
+    def test_find_event_with_project_filtering(self):
```
mikkonie: Again, I recommend splitting this into multiple tests.

docs/source/major_changes.rst
```diff
@@ -38,6 +38,16 @@ access the ``user_modifiable`` variable of an app setting, you should instead
...
 
+Plugin API search() Changes
+---------------------------
+
+The ``search()`` function implemented by plugins already had the ``keywords``
+argument in its signature, but so far it was not used. This release implemented
+support for the ``project:uuid`` keyword. All SODAR Core apps are now expected
+to support this keyword as well, which, if present in the ``keywords``
+dictionary, should restrict the search to the project or category specified by
+the UUID.
```
mikkonie: This should be updated to reflect that the signature of `search()` has changed, and the `projects` kwarg contains the project queryset and must be supported.

mikkonie (changes requested):
Excellent work. I was especially impressed with the thorough testing.

I also played around with the feature in dev and wasn't able to break anything despite my best efforts :)

I have some comments, but this is all minor cleanup stuff.

filesfolders/tests/test_models.py
fmarotta: I removed it to reduce the redundancy.

filesfolders/tests/test_models.py
fmarotta: Alright, sounds like a good convention :)

filesfolders/tests/test_models.py
fmarotta: Done

filesfolders/models.py
fmarotta: You are right, I probably copied it from the timeline code without checking.

projectroles/tests/test_models.py
fmarotta: Done

projectroles/tests/test_models.py
fmarotta: Done

projectroles/tests/test_views.py
fmarotta: OK, good to know

projectroles/tests/test_views.py
fmarotta: Also done

timeline/tests/test_models.py
fmarotta: Done!

projectroles/models.py
fmarotta: Done

docs/source/major_changes.rst
fmarotta: Ah right, I forgot to update the change log, thanks for spotting this.

fmarotta:
Thanks for the review, perhaps spending time on the tests wasn't such a smart idea since now we are about to refactor the search API completely, but hopefully it won't be too painful to update the tests at this point :)

mikkonie:
> Thanks for the review, perhaps spending time on the tests wasn't such a smart idea since now we are about to refactor the search API completely, but hopefully it won't be too painful to update the tests at this point :)

We might still figure out that some breaking changes to the API might not be work as we envisioned, so it's perfectly fine to iterate the tests as well :)

mikkonie (approved):
Thanks for the fast work going through my comments. Approved, will merge :)

Issue #1899, opened 2026-03-03
Provide pre-filtered Project queryset to plugins in search()


mikkonie:
Implementations for #1191 and #1898 would give a lot of project filtering responsibility to app plugins. Basically, all implementers of `search()` must do their own filtering for projects corresponding to the keyword parameters as well as user access. This increases the possibility of errors in implementation, as well as potential overhead for redundant Django database queries.

Apps do need to do some degree of project filtering anyway to e.g. filter for a specific permission (see filesfolders for an example). Nevertheless, requiring all apps to do their own filtering for the `project` keyword from ground up makes little sense, especially if we want to filter for partial title matches and such.

We should considering adding a `projects` argument to `ProjectAppPlugin.search()`. This argument would contain a `QuerySet` of `Project` objects already filtered down to the following:

- Projects for which the user has any role (local or inherited)
- Projects filtered down to a possible project UUID or (partial) project title via the `project` keyword.
- (+ possible future "global" project limiting we may come up with, without requiring changes in apps)

This would mean plugins implementing search would not have to worry about having to limit their queries to specific projects, instead simply passing the pre-filtered queryset to their own object filters. The plugin search implementations are, of course, still free to do further filtering down to specific projects when needed.

An alternative approach would be to filter results down in projectroles when we receive them from the plugins. Alas, this would potentially be very inefficient, as apps may need to return objects as lists instead of querysets, or query for something that's not even in the Django model (e.g. SODAR iRODS searches). This would mean the plugins querying and processing redundant data which then gets discarded by projectroles. Hence I don't believe this is the way to go.

Possible challenges and downsides:

- This will be a breaking change. No nice way to do it via depecation either, unless we want to create an alternative `search()` function and deprecate the old one.
- For efficiency, we need to provide the projects as `QuerySet`, not a `list`. We can't simply iterate through an array of already queries and call `has_role()` for each to catch all the inherited roles. That would perform badly anyway. However, there should be ways around this. Similar project filtering is already done for the project list, we could start looking there..

Comments are very much welcome. If we decide to go with this, we'd probably want to rework the pull request for #1191 right away..

mikkonie:
We discussed this with @fmarotta and agreed to implement it in iterations:

1. Add this in the PR for project keyword feature (#1191), initially only supporting project UUID
2. Implement project partial title filtering in (#1898), filter projects accordingly for the arg there
3. *Possibly* implement pre-filtering the project queryset by user roles in projectroles.

[...]

Issue #1898, opened 2026-03-03
Add title support for project keyword in search


mikkonie:
Suggested by @fmarotta, related to #1191. It might make sense to allow providing a (partial) project title in the `project` keyword for `search()`.

The average user might expect to be able to limit the search by project title. As only full titles are unique and can get very long in real life, this should search for partial matches and filter accordingly.

This would require the ability to define terms with whitespace limited by quotes, e.g. `project:"partial project title here"`. Do we support that yet? I forget.

Let's do this *after* we have merged #1191 to keep things nice and iterative :)

## Spec

- Ensure we can support keywords with whitespace and quotes as limiters
   * If not, add support and tests
- Add support for partial project title search when parsing the `project` keyword
   * IF keyword doesn't correspond to a UUID
   * THEN filter the project pre-filtering queryset (developed in #1899) down to `full_title__icontains=keyword`
   * ELSE filter down to UUID as we do in #1191
- Add tests to ensure we receive expected projects
- Update documentation

mikkonie:
Done by @fmarotta, merged.

PR #1903, opened 2026-03-09
Add title support for project keyword in search

fmarotta:
This PR attempts to solve #1898 and address the second point of [this comment](https://github.com/bihealth/sodar-core/issues/1899#issuecomment-4005604798) in #1899. I used the `shlex` module to parse the search string in a way that preserves the quotes. As a side effect, search terms can now also be quoted. I also updated the docs about the signature for the search() function.

### DIFF
--- a/projectroles/tests/test_views.py
+++ b/projectroles/tests/test_views.py
@@ -442,6 +442,27 @@ class TestProjectSearchResultsView(
             len([p for p in self.plugins if p.search_enable]),
         )
 
+    def test_get_quoted(self):
+        """Test ProjectSearchResultsView GET with quoted search term"""
+        with self.login(self.user):
+            response = self.client.get(
+                reverse('projectroles:search')
+                + '?'
+                + urlencode({'s': '"test quoted"'})
+            )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['search_terms'], ['test quoted'])
+        self.assertEqual(response.context['search_keywords'], {})
+        self.assertEqual(response.context['search_type'], None)
+        self.assertQuerySetEqual(
+            response.context['search_projects'], Project.objects.all()
+        )
+        self.assertEqual(response.context['search_input'], '"test quoted"')
+        self.assertEqual(
+            len(response.context['app_results']),
+            len([p for p in self.plugins if p.search_enable]),
+        )
+
     def test_get_search_type(self):
         """Test GET with search type"""
         with self.login(self.user):
@@ -487,6 +508,10 @@ class TestProjectSearchResultsView(
             {'project': str(self.project.sodar_uuid)},
         )
         self.assertEqual(response.context['search_type'], None)
+        self.assertQuerySetEqual(
+            response.context['search_projects'],
+            Project.objects.filter(sodar_uuid=self.project.sodar_uuid),
+        )
         self.assertEqual(
             response.context['search_input'],
             f'test project:{self.project.sodar_uuid}',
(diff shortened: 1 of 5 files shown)


projectroles/tests/test_views.py
```diff
@@ -442,6 +442,27 @@ def test_get(self):
             len([p for p in self.plugins if p.search_enable]),
         )
 
+    def test_get_quoted(self):
+        """Test ProjectSearchResultsView GET with quoted search term"""
```
mikkonie: Minor note on conventions: since this isn't the first GET test for this view, we can simply write `Test GET...` in the docstring.

projectroles/tests/test_views.py
```diff
@@ -748,6 +917,21 @@ def test_get_app_search_results_within_project(self):
...
 
+    def test_get_app_search_results_within_project_by_title(self):
+        """Test GET app search results within project by title"""
+        with self.login(self.user):
+            response = self.client.get(
+                reverse('projectroles:search')
+                + '?'
+                + urlencode({'s': 'test_event project:"{self.project.title}"'})
+            )
+        # Events are not found when search is restricted to `self.project`
```
mikkonie: Small cleanup thing: the quotes are probably not needed here :)

projectroles/tests/test_views.py
```diff
@@ -497,31 +522,146 @@ def test_get_keywords(self):
...
+            response.context['search_input'],
+            f'test project:"{self.category.title}"',
+        )
+        self.assertEqual(
+            len(response.context['app_results']),
+            len([p for p in self.plugins if (p.search_enable)]),
         )
-        self.assertEqual(len(response.context['project_results']), 0)
+
+    def test_get_search_string_parsing(self):
```
mikkonie, line 613: Nothing to correct here, just wanted to say that this is very nicely tested! You're more thorough than I would've been :)

projectroles/views.py
```diff
@@ -869,20 +871,18 @@ def get_context_data(self, *args, **kwargs):
                 for t in self.request.POST['m'].strip().split('\r\n')
                 if len(t.strip()) >= 3
             ]
+            search_input = shlex.join(search_terms)
```
mikkonie: I noticed one issue when using the advanced search.

If I enter the same string quoted and un-quoted, it seems we don't correctly detect these as dupes. If I do the same thing by either quoting or un-quoting both duplicate terms, the dupe is successfully removed. See screenshots:

Image

Image

Furthermore, maybe we'd like to strip the quotes from when we display the terms in the template? Now we end up with double quotes which looks a bit funny.

mikkonie (changes requested):
Solid work. Again, very nicely tested! I did find one minor bug, otherwise just small cleanup notes.

projectroles/tests/test_views.py
fmarotta: OK, good to know

projectroles/tests/test_views.py
fmarotta: Removed

projectroles/tests/test_views.py
fmarotta: Thank you, while I'm at it I will also add a test for the bug you found :)

projectroles/views.py
fmarotta: Good catch, the terms for advanced search were not parsed in the same way as the ones for standard search. Should be fixed now.

mikkonie: I can confirm the results page title is now fixed. However, when I repeated the input from above, I noticed the following:

2026-03_search_quote_bug3

I don't think this was present in the previous commit I reviewed, but it's possible I just didn't notice it..

fmarotta: I restored the original behaviour of showing an empty bar in the advanced search results.

mikkonie (approved):
Looking good! Approved, will merge.

Issue #1881, opened 2026-02-03
Replace Black and Flake8 with Ruff


fmarotta:
We could consider using ruff to replace black.

Advantages:
- The 'single quotes' code convention could be enforced automatically
- Sorting imports alphabetically could also be enforced, including custom sections (e.g., for django and projectroles imports)
- Code formatting would be faster
- Some of flake8's features could also be covered

Disadvantages:
- We would need to spend some time to understand the configuration options
- Despite the claims to be a drop-in replacement, in some cases ruff is not completely equivalent to black and flake8

I have run some tests (just for the single quotes feature):
- `pip install ruff`
- Apply the following patch
```
diff --git a/pyproject.toml b/pyproject.toml
index a36380ab..a8ee9125 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,9 @@
 requires = ["setuptools>=80.9.0", "versioneer[toml]==0.29"]
 build-backend = "setuptools.build_meta"
 
-[tool.black]
+[tool.ruff]
 line-length = 80
-skip-string-normalization = true
-exclude = ".git|.venv|.tox|build|env|src|docs|versioneer.py"
+exclude = [".git", ".venv", ".tox", "build", "env", "src", "docs", "versioneer.py"]
+
+[tool.ruff.format]
+quote-style = "single"
```

Some observations:
- Single quotes are indeed enforced
- If the string contains an apostrophe, ruff uses double quotes instead
- Ruff does use a blank line after a class docstring, but if the class doesn't have a docstring, it doesn't put an empty line before the class fields.

mikkonie:
This seems like an idea worth considering, especially as (if I understood correctly) you've already had positive experiences with the package.

I was under the impression this needed some Rust-specific dependencies but apparently not, can be just installed via PyPI?

I also appreciate the experimenting a lot. I'll do some playing around of my own and see if the introduced differences to black are an issue. As a single-quote supporter, I very much appreciate the fact that we can enforce those :)

fmarotta:
Yes, it can be just installed with pip from PyPI. My experience with it was positive, although mostly limited to the default configuration :)

mikkonie:
Black v26.3.1 has a fix for a security issue, which apparently affects all prior versions. Upgrading to that version and reformatting results in a LOT of changes to our files, apparently because of a single thing which has changed in Black's heuristics.

[...]

PR #1905, opened 2026-03-13
Replace Black and Flake8 with Ruff (#1881)

mikkonie:
Setting up Ruff to replace Black and Flake8 for formatting and linting. This intends to handle #1881.

I did the following:

- Replace dependencies
- Set up configuration for linting and formatting in `pyproject.toml`
- Reorganize `Makefile` with separate commands for linting, formatting and checking
    * These don't have to be used, but it provides a consistent interface if e.g. a better option than Ruff emerges in the future :)
- Update CI workflows
- Update docs

**Note:** At the time of writing, reformatting the code has **NOT** yet been done in this work branch. I'll do that once I'm ready to merge, in order to avoid unnecessary merge conflicts with other ongoing PRs. @fmarotta, let's discuss when would be the best time to do this :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,9 +42,7 @@ collectstatic:
 
 
 .PHONY: check
-check:
-	ruff format --check
-	ruff check $(arg)
+check: format-check lint
 
 
 .PHONY: js-beautify
--- a/Makefile
+++ b/Makefile
@@ -3,16 +3,17 @@ MANAGE = python manage.py
 define USAGE=
 @echo -e
 @echo -e "Usage:"
-@echo -e "\tmake black [arg=--<arg>]                 -- black formatting"
+@echo -e "\tmake black [arg=--<arg>]                 -- format Python with black (to be removed)"
+@echo -e "\tmake celery                              -- start celery worker"
+@echo -e "\tmake collectstatic                       -- run collectstatic"
 @echo -e "\tmake flake                               -- run flake8"
 @echo -e "\tmake js-beautify arg=<path>              -- run js-beautify on Javascript file(s)"
-@echo -e "\tmake celery                              -- start celery worker"
+@echo -e "\tmake manage_target arg=<target_command>  -- run management command on target site, arg is mandatory"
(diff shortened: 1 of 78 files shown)


Makefile
```diff
@@ -28,39 +31,25 @@ black:
...
 .PHONY: collectstatic
 collectstatic:
 	$(MANAGE) collectstatic --no-input
 
 
-.PHONY: test
-test: collectstatic
-	$(MANAGE) test -v 2 --parallel --settings=config.settings.test $(arg)
+.PHONY: check
+check:
```
fmarotta: Here we could avoid duplication by writing
```
check: format-check lint
```
and having an empty rule body.

fmarotta:
Everything looks good to me, I just had one minor comment for the Makefile.

On a tangential note, since we are also touching the CI workflow, I think it would make sense to run the formatting and linting before the tests. It's a bit disheartening to wait 30 minutes for the tests to pass only to find out that the build fails due to a formatting issue.. Moreover, after fixing the format, the tests will have to run again. Normally it shouldn't happen because we always follow the good practices and check everything locally before pushing, so it's a very minor thing. But it happened to me a few times, I must admit. Curious to know what you think about it.

Makefile
mikkonie: Good point! I will update.

mikkonie:
> On a tangential note, since we are also touching the CI workflow, I think it would make sense to run the formatting and linting before the tests. It's a bit disheartening to wait 30 minutes for the tests to pass only to find out that the build fails due to a formatting issue.. Moreover, after fixing the format, the tests will have to run again. Normally it shouldn't happen because we always follow the good practices and check everything locally before pushing, so it's a very minor thing. But it happened to me a few times, I must admit. Curious to know what you think about it.

This is a good point. Not sure why we have decided to do it this way originally. The 30 minutes in SODAR Core is even quite short, CI in SODAR takes 60min+. I'll add a separate ticket for this for easier follow-up.

fmarotta (approved):

Issue #1902, opened 2026-03-09
Replace search() "keywords" kwarg with **kwargs


mikkonie:
I would like to think this is more Pythonic :) Not sure why it wasn't implemented like this in the first place? My bad.

This is a breaking change for the project app plugin search API and needs to be documented as such.

## Spec

- Replace `search()` `keywords` kwarg with `**kwargs`
- Update implementations to parse keywords from possible optional kwargs
- Update tests
- Update documentation for breaking changes in `search()` signature

mikkonie:
Done by @fmarotta, merged.

Issue #1901, opened 2026-03-09
Replace search() "search_type" kwarg with "keywords" item


mikkonie:
Since we are already blowing up the project app plugin search API with the addition of the `projects` kwarg, it would be a good time to rip off the band-aid and remove the `search_type` kwarg from the method signature.

The search type should be provided in `keywords`, as it is a simple key-value pair. It should be treated like any other (potential) keyword instead of its own special case.

Existing implementations of `search()` in `filefolders` and `timeline` should be modified to reflect this, of course. This should also be clearly documented as a breaking change for v1.4.

## Spec

- Parse `search_type` into `search_keywords` in `ProjectSearchResults.get_context_data()`
- Handle parsed search type accordingly in `ProjectSearchMixin`
- Update `ProjectAppPlugin.search()` to not include the separate `search_type` kwarg
- Update `search()` implementations
    * Omit the removed kwarg
    * Use `type` from `keywords`
- Update tests
- Document breaking changes

## Tasks

- [ ] Update search type parsing in projectroles
- [ ] Update `search()` signature
- [ ] Update implementations
- [ ] Update tests
- [ ] Document

mikkonie:
Done by @fmarotta, merged.

PR #1906, opened 2026-03-16
Replace search() "search_type" kwarg with "keywords" item and "keywords" kwarg with **kwargs

fmarotta:
This PR addresses #1901 and #1902. It should be rebased after #1903 is merged. I followed the spec pretty closely. In addition, I modified the templates as necessary. I wasn't sure whether to use `**kwargs` also for the find() methods in the model managers.


mikkonie (approved):
Good work, no notes! I'll merge once CI passes.

Issue #1916, opened 2026-03-26
Enable AdminAlertListView viewing for non-superusers


mikkonie:
While discussing admin alert related issues with a user, I realized that we still haven't implemented the ability for users to dismiss admin alerts.

Before we do that, we should enable the user to see the admin alert list. Otherwise, dismissed alerts will be lost forever (if not emailed). This is the ticket to achieve that.

Needless to say, a regular user should not be able to edit or delete the alerts, simply view them.

## Spec

- Change permission for `adminalerts.view_list` into `rules.is_authenticated`
- Update `alert_list.html` layout for non-superusers
    * Hide "Create Alert" button
    * Hide update/delete dropdowns
    * Show status, but do not allow user to click on button to modify it
- Update tests
    * Update permission tests to assert regular user access
    * Update UI tests to assert editing controls are not rendered for regular user
- Update docs
    * In `app_adminalerts.html` under "Usage", we should make clear that all authenticated users can view the list, while only superusers can create and edit alerts

## Tasks

- [x] Update view permissions
- [x] Update view layout
- [x] Update tests
- [x] Update docs

mikkonie:
Done and merged.

PR #1919, opened 2026-03-31
Enable AdminAlertlistView viewing for non-superusers

fmarotta:
This PR addresses #1916 by implementing the spec described therein.

### DIFF
--- a/adminalerts/templates/adminalerts/alert_list.html
+++ b/adminalerts/templates/adminalerts/alert_list.html
@@ -47,10 +47,10 @@
 <div class="row sodar-subtitle-container bg-white sticky-top">
   <h2><i class="iconify" data-icon="mdi:alert"></i> Admin Alerts</h2>
   {% if request.user.is_superuser %}
-  <a role="button" class="btn btn-primary ml-auto" id="sodar-aa-alert-create"
-     href="{% url 'adminalerts:create' %}">
-    <i class="iconify" data-icon="mdi:plus-bold"></i> Create Alert
-  </a>
+    <a role="button" class="btn btn-primary ml-auto" id="sodar-aa-alert-create"
+       href="{% url 'adminalerts:create' %}">
+      <i class="iconify" data-icon="mdi:plus-bold"></i> Create Alert
+    </a>
   {% endif %}
 </div>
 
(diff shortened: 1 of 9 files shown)


adminalerts/templates/adminalerts/alert_list.html
```diff
@@ -46,10 +46,12 @@
 
 <div class="row sodar-subtitle-container bg-white sticky-top">
   <h2><i class="iconify" data-icon="mdi:alert"></i> Admin Alerts</h2>
-  <a role="button" class="btn btn-primary ml-auto"
+  {% if request.user.is_superuser %}
+  <a role="button" class="btn btn-primary ml-auto" id="sodar-aa-alert-create"
```
mikkonie: Minor cleanup note: make sure to indent this block inside the `if` clause.

adminalerts/templates/adminalerts/alert_list.html
```diff
@@ -83,7 +85,8 @@ <h2><i class="iconify" data-icon="mdi:alert"></i> Admin Alerts</h2>
                 <button class="btn sodar-list-btn
                   {% if not a.is_expired and a.active %}btn-success
                   {% elif not a.is_expired and not a.active %} btn-warning
-                  {% else %} btn-danger{% endif %} js-change-alert-state-button"
+                  {% else %} btn-danger{% endif %}
+                  {% if request.user.is_superuser %} js-change-alert-state-button {% endif %}"
```
mikkonie, line 89: As regular user, the button still appears to be clickable. We should explicitly disable it as a button in this case, by e.g. adding a `disabled` attribute to it.

mikkonie (changes requested):
Good work. I noticed one issue, but I may have forgotten to write about this in the spec:

When logging in as a non-superuser, the adminalerts app is not visible in the user dropdown. I believe this should be fixable by changing the following in `adminalerts.plugins.SiteAppPlugin`:

```
app_permission = 'adminalerts.create_alert'
```

..to:

```
app_permission = 'adminalerts.view_alerts'
```

adminalerts/templates/adminalerts/alert_list.html
fmarotta: Done

adminalerts/templates/adminalerts/alert_list.html
fmarotta: Done

fmarotta:
Thanks, I missed that. Should be fixed, and I've added an extra UI test.

mikkonie (approved):
Excellent! No further notes, will merge once CI passes :)

Issue #1918, opened 2026-03-26
Update site app messages to support custom dismissal Ajax view URL


mikkonie:
We want to enable dismissal of admin alerts along with storage of dismissal status per user. For that, see #1917.

Alas, the way these alerts are implemented possibly makes this non-trivial.

Adminalerts is a separate plugin-based app, which may be enabled on a site or not. Hence we can't just hardcode displaying these alerts in the site template. For this, I have done the "site app messages" framework, which allows displaying messages from site apps by implementing `SiteAppPluginPoint.get_messages()`.

How this currently works is as follows:

- Projectroles calls `get_messages()` in site app plugins
- Plugins return a list of dicts with alert content and additional fields
    * `color` = Alert color variant
    * `dismissable` = Boolean, true for dismissable
- Messages are retrieved server-side within the `_messages.html` template using the `get_site_app_messages()` template tag.
- The `dismissable` field simply adds the Bootstrap classes `alert-dismissable fade show`.
   * ..which means it's shown again after reload, right?
   * It doesn't seem this feature is very well thought out..

What we basically need to do is:

- Create a model to store alert dismissals per user
   * Basically just containing alert object foreign key and user object foreign key
- Create an Ajax view to submit dismissals by creating an object with the modal
- Link the site app dismissals to the Ajax view with the correct URL.

We don't want to spend a lot of time making complex changes to this API, as it's not used much and might eventually be replaced with something much better anyway.

After thinking about this for some time, I think the easiest solution is as follows:

- Add support for declaring `dismiss_url` field in messages
- Place `dismiss_url` in a `data` attribute of the message alert element in the template
- In `projectroles.js`, add JQuery for making a request to `dismiss_url` when user clicks on dismissal, if the URL is present
- Implement the rest within the adminalerts app in #1917.

## Spec

- Update `_messages.html` template
    * Add `message.dismiss_url` to `data-dismiss-url` attribute
- Add JQuery event listener
    * IF dismiss button in `.sodar-alert-site-app` is called
    * AND `data-dismiss-url` has a value
    * THEN perform a `GET` request to url
        + This can be async, e.g. we can just submit request and close the alert without waiting for a response
        + Log a non-200 response in the console
- Add `dismiss_url` in `get_messages()` docstring example
    * Mention this is optional
- Update tests if needed
    * I think we can leave the actualy testing of this feature in #1917.
- Update docs if needed

## Tasks

- [x] Update template
- [x] Add JQuery event listener
[...]

Issue #1917, opened 2026-03-26
Add admin alert dismissal


mikkonie:
This should be done after #1916 and #1918, or together with #1918 if that seems viable.

Right now, users are not able to dismiss admin alerts. This can be annoying, as hinted to me by several real world users :)

Allowing the dismissal of admin alerts also allows admins to be a bit more liberal with creating them, which is nice. We should not be afraid of informing users of ongoing issues.

We should have the site remember dismissals, so a Django database model for dismissed alerts needs to be created and used.

Since these alerts are dynamically retrieved by the "site app messages" framework, changes to that are needed first. That will require a separate ticket, to be linked here.

## Spec

- Add `AdminAlertDismissal` model
    * `alert`: `ForeignKey` to `AdminAlert`
    * `user`: `ForeignKey` to `User`
    * `modified`: Autofill datetime for modification
- Add `AdminAlertDismissAjaxView`
    * Support `GET` request
    * URL = `ajax/dismiss/<uuid:adminalert>`
        + User will be retrieved from request via session
    * If user is not authenticated, do nothing
    * Otherwise create `AdminAlertDismissal` object for alert and user
        + Maybe use `get_or_create()` in case we somehow try to do a double-dismissal
- Pass URL to alert dismissal in `get_messages()` as `dismiss_url`
- Update AdminAlert querying in `get_messages()`
    * Don't return alerts already dismissed by user
    * We should try to optimize required queries as much as possible to not slow down the page load on large sites, but we can iterate this if needed
- Add/update tests
    * Permission and view tests for the new Ajax view
    * Possible plugin tests for `get_messages()`?
- Update docs where relevant

## Tasks

- [x] Add model
- [x] Add Ajax view
- [x] Update `get_messages()`
- [x] Update tests
- [x] Update docs

mikkonie:
Done by @fmarotta and merged.

PR #1920, opened 2026-04-01
Add admin alert dismissal

fmarotta:
This PR addresses #1917 and #1918, adding full support for dismissing admin alerts. Here is a summary of what was done:

- Add a model for `AdminAlertDismissal`, containing user-alert pairs representing individual dismissals;
  - I also add a `dismissed_by` many-to-many field in the `AdminAlert` model, as I find that it can simplify the query in `get_messages()`;
  - Generate the corresponding migration;
- Update `adminalerts.plugins.get_messages()`;
  - Only return alerts which were not dismissed by the user;
  - Add `dismiss_url` field in the returned dictionary;
  - Set the `dismissable` field to `True`;
- Add `AdminAlertDismissAjaxView`;
  - It supports GET requests to `get_or_create()` dismissal entries;
- Modify the `_messges.html` template to allow users to dismiss alerts.
- Add tests
- Update docs
- Manually check in the local dev instance

Comments for possible future iterations:
- When anonymous users request the ajax view, nothing happens and the view responds with 200. I think it could make sense to also require `adminalert.view_alert` permission in this view, so that the request would return 403. Maybe just returning 401 would also make sense. EDIT: on second thought, that would leave an error in the browser's console log, and could be confusing.

Last comment: I also updated the doc page for appalerts, fixing what I suspect is a typo (it was referencing the adminalerts config).


adminalerts/plugins.py
```diff
@@ -114,8 +120,13 @@ def get_messages(self, user: Optional[User] = None) -> list[dict]:
                 {
                     'content': content,
                     'color': 'info',
-                    'dismissable': False,
+                    'dismissable': True,
```
mikkonie: This should only be dismissable when the user is authenticated, so that an anonymous user does not get the dismiss link on the alerts. Currently the dismiss link is visible:

Image

It's possible I forgot to write this in my spec and in that case, sorry about that :)

Also, make sure to add UI tests to assert dismiss link visibility according to user auth status.

projectroles/templates/projectroles/_messages.html
```diff
@@ -33,6 +33,11 @@
         <div class="alert alert-{{ message.color }} {% if message.dismissable %}alert-dismissable fade show{% endif %} sodar-alert-top sodar-alert-site-app">
           <div class="sodar-alert-top-content">
             {{ message.content | safe }}
+            {% if message.dismiss_url %}
+              <a href="#" data-dismiss="alert" class="pull-right sodar-alert-close-link" data-dismiss-url="{{message.dismiss_url}}">
```
mikkonie: I guess this wasn't in my original spec, but we should copy the fade-out transition we use in dismissing regular Django alerts here, just to keep things nicely consistent.

adminalerts/views_ajax.py
```diff
@@ -25,3 +27,22 @@ def post(self, request, **kwargs):
         alert.active = not alert.active
         alert.save()
         return Response({'is_active': alert.active}, status=200)
+
+
+class AdminAlertDismissAjaxView(APIView):
```
mikkonie: To answer your question in the PR comments: yes, this view should be access controlled by perms in `rules.py`. An anonymous user should not be making requests here anyway, as they should not have the dismissal link available in the UI (see my comment elsewhere).

I recommend adding a new `dismiss_alert` perm where we check for `user.is_authenticated`.

adminalerts/migrations/0008_adminalertdismissal_adminalert_dismissed_by.py
```diff
@@ -0,0 +1,63 @@
...
+                    models.ForeignKey(
+                        help_text='User who has dismissed the alert',
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name='dismissed_alerts',
+                        to=settings.AUTH_USER_MODEL,
+                    ),
+                ),
+            ],
+        ),
+        migrations.AddField(
```
mikkonie: `ManyToManyField` can be quite tricky to handle and IMHO should be avoided whenever possible. Especially when introducing it to models between different apps, or from dynamic apps into non-app models such `User`. What happens if the developers of a site decide to disable or remove a previously included adminalerts app?

Instead, I recommend the following:

1. Add `related_name='dismissals'` to `AdminAlertDismissal.alert`
2. Query with `alerts.exclude(dismissals__user=current_user_here)`

This should work and be performant, unless there is some case of which I'm not thinking right now. Let me know if you think of an even better solution.

projectroles/templates/projectroles/_messages.html
```diff
@@ -51,3 +56,17 @@
     {% endfor %}
   </div>
 {% endif %}
+
+<script>
+  $('.sodar-alert-site-app .sodar-alert-close-link').each(function () {
+    this.addEventListener("click", function (event) {
```
mikkonie: Double quotes! :)

projectroles/templates/projectroles/_messages.html
mikkonie: Also, super minor cleanup note here: Let's maintain the formatting of spaces between `{{ }}` tags and their content here for maxium searchability.

mikkonie (changes requested):
Good work as always, works as intended. No bugs discovered despite my best efforts :)

I have some requests for improvements and tweaks.

adminalerts/plugins.py
fmarotta: It makes sense, I fixed it.

projectroles/templates/projectroles/_messages.html
fmarotta: That should already be the case, or am I missing something?

adminalerts/views_ajax.py
fmarotta: Done.

adminalerts/migrations/0008_adminalertdismissal_adminalert_dismissed_by.py
fmarotta: Sounds good, and it looks like the query is exactly the same.

projectroles/templates/projectroles/_messages.html
fmarotta: Ahhh, fixed.

projectroles/templates/projectroles/_messages.html
fmarotta: Done

fmarotta:
Thanks for the thorough review as always :) I have addressed the comments.

projectroles/templates/projectroles/_messages.html
mikkonie: Interesting, it seems the fadeout has been removed from regular alerts as well. Intentional or a bug, can't remember.

In any case, it does appear the admin alert dismissal does behave identically to other Django alerts. Please disregard and sorry for the false alarm :)

mikkonie (approved):
Looks good, works great, no notes. Will merge :)

Issue #22, opened 2018-08-24
Add client-side retrieval of UI search results


mikkonie:
**Update 2026-03:** I'm reviving this old issue and cleaning up old comments.

AT the moment, UI search and results displaying is performed 100% on the server side. This does not work great for all real-life use cases, as SODAR Core can't control the performance or waiting time needed for different apps to produce results.

Instead, we should have client-side JQuery perform plugin-specific queries for apps and display results for each app as they arrive.

Some considerations:

- We should maintain a specific ordering of the apps/plugins for each query, regardless of if and when they return results.
- We should clearly display a "no results" or "error querying for results" element for each queried plugin, maintaining the order mentioned in the previous point.
- DataTables needs to be set up for each plugin's result set, as it is in the current implementation.
- Breaking changes are needed for apps' `search()` return data (see below)

Problems with current method of app search results:

- Currently, apps are free to determine rendering the results table however they want
- With client-side retrieval, we can't do this anymore
    + ..unless we have the search API return HTML, which is obviously ill-advised

Solution:
 
- We meed to declare a standardized return data format for `search()`
    * Something to allow us to build the DataTable client-side
    * ..while supporting as many of the custom table rendering things currently done in the app result templates as possible
- This also needs to be nicely serializable into Ajax view return data for the client-side JQuery
- The basic sctructure would be something like:
    * List of table columns with column names and potential attributes regarding e.g. formatting
    * List of rows of data for the results table
- **NOTE:** One plugin may return multiple results tables depending on search type!
    * See e.g. samplesheets app in SODAR
    * To keep it simple, let's query for all results from the plugin in one request and then display 1-N tables depending on what the plugin returns

### Update ProjectSearchResultsView for Client-Side Search

- Add dict of app plugin search types to be searched in context data
    * Return list of `ProjectAppPlugin`:s with `search_enable=True`
    * Make sure to order this alphabetically by plugin name
    * For each plugin, include `plugin_name`, `plugin_title` and `icon`
- Disable server-side search results retrieval in the view
    * Don't delete the code yet, we'll have use for elsewhere :)
- (This might be a good time to update tests to remove expected server-side search results)

### Update search API and Implementations

- Update expected return data from project app plugin `search()` methods
    * `search()` should return 1-N result tables
        + Column & formatting info
        + Rows with the actual results
    * Require `search()` to provide results in a specific Python object format
        + Ajax view will serialize this to JSON for the client
    * **TBD:** Specific format
[...]

PR #1922, opened 2026-04-08
Add Ajax retrieval of search results

fmarotta:
## Summary

This PR attempts to address #22.

Changes in `projectroles/plugins.py`:
* Added `PluginSearchResultCell` class to represent a generic cell in search results tables
* Added `PluginSearchResultColumn` class to store column information
* Modified `PluginSearchResult` to contain a list of `PluginSearchResultColumn` and a list of lists of `PluginSearchResultCell`s

Changes in `projectroles/views.py`:
* `ProjectSearchResultsView` now doesn't perform the search, but simply returns a template with placeholders for the plugins which should be searched
* Plugin-specific CSS files are also included in this template
* The template relies on JavaScript functions in `projectroles/static/projectroles/js/search.js`, which collect the search results and add them to the HTML of the page

Changes in `projectroles/views_ajax.py`:
* Added `ProjectSearchResultsAjaxView` which can be called with POST requests for each plugin individually. It performs the search and returns the results serialized as JSON.

Changes in `timeline/plugins.py` and `filesfolders/plugins.py`:
* Updated the search() method to return results using the new `PluginSearchResult`, `PluginSearchResultColumn`, and `PluginSearchResultCell` classes
* Removed the search-related templates since now results tables are built with JavaScript

Changes in `projectroles/js/projectroles/search.js`:
* Added functions to perform the search asynchronously, build the results tables, and highlight terms

Changes to template tags:
* Update the signature of `get_remote_icon()`
* Replace `timeline.get_app_icon_html()` with `timeline.get_app_badge()` which is more general and can be used both in search results and in `_list_item.html`. As a side effect, the link applies to the text, not just the icon, which should make this badge more consistent with the others.

~Changes to CSS classes:~
~* `.sodar-overflow-container` should now be applied to the parent `<TD>` tag instead of the inner `<DIV>`. The CSS and JS work under this assumption now. The reason is that the timeline plugin search table has a dynamic cell class which changes the background color based on the timeline event status; this should be applied to the `<TD>`. It seemed more economical to assign the dynamic class to the whole `<TD>`, but let me know if you have better ideas.~

## Notes

* [x] I have not updated the UI tests, pending decisions about the final structure of the implementation.
* [x] I have also postponed checking the Tour
* [x] Will also need to add tests for `result_limit`
* [x] I will go through the docs again and update them


projectroles/views.py
```diff
@@ -959,6 +833,7 @@ class ProjectAdvancedSearchView(
     """View for displaying advanced search form"""
 
     template_name = 'projectroles/search_advanced.html'
+    http_method_names = ['post']
```
mikkonie: This change makes the advanced search UI unavailable to the user. When trying to open the view I get `HTTP 405` because `GET` is no longer approved.

projectroles/static/projectroles/js/search.js
```diff
@@ -1,67 +1,273 @@
...
+            class: 'form-control sodar-search-page-length'
+          }).append(
+            $('<option>', {
+              value: window.searchPagination,
+              selected: true
+            }).text('Page'),
+            $('<option>', {
+              value: window.searchPagination
+            }).text(`${window.searchPagination} (default)`),
+            $('<option>', {
```
mikkonie: This might be cleaner if we create a constant for the options and then iterate on that to populate the select field.

projectroles/static/projectroles/js/search.js
```diff
@@ -1,67 +1,273 @@
...
+          $('<input>', {
+            class: 'form-control sodar-search-filter',
+            type: 'text',
+            placeholder: 'Filter',
+            ariaLabel: `Filter ${cardTitle}`
+          }),
+        ),
+      ),
+    ),
+    // XXX: To be discussed: I found this code in the old _search_header.html template, but I can't see any other reference to result_limit. Is it still relevant?
```
mikkonie: This is used in SODAR, so we should support it.

projectroles/views_ajax.py
```diff
@@ -490,6 +500,209 @@ def post(self, request, *args, **kwargs):
...
+        plugin = plugin_api.get_app_plugin(plugin_name)
+        if search_type is not None and search_type not in plugin.search_types:
+            return {
+                'error': f'The app "{plugin_name}" does not support search results of type "{search_type}".',
+                'results': None,
+            }
+        try:
+            results = plugin.search(terms, user, projects, **keywords)
+            # TODO: The original code organized results by category, so results should be a dict where the categories are the keys.
+            # XXX: To be discussed: how should we group results by category?
```
mikkonie: It is OK to simply return these as a list of serialized `PluginSearchResult` objects, as is done right now.

The keys are redundant, as the JQuery can simply parse the result objects into one or multiple tables containing the results. All the information needed for rendering each table should be provided within the `PluginSearchResult` object.

timeline/templatetags/timeline_tags.py
```diff
@@ -136,6 +136,54 @@ def get_app_icon_html(event: TimelineEvent, plugin_lookup: dict) -> str:
     )
 
 
+@register.simple_tag
+def get_app_badge(
```
mikkonie, line 95: This is defined as a template tag, but it's not used in any templates.

If this is intended for purely to be used in plugins for search, IMHO it would make more sense to place it into the plugin class as an internal helper.

..but writing this, I realize this also ties to your question in the PR about how to implement this. I'll have to think about it a bit and get back to you.

projectroles/templates/projectroles/search_results.html
```diff
@@ -96,56 +91,24 @@ <h2 class="sodar-pr-content-title pt-auto">Search Results</h2>
...
-      {% if app.plugin.search_template and app.has_results and not app.error %}
-        {% include app.plugin.search_template with plugin=app.plugin search_results=app.results %}
-      {% elif app.error %}
-        <div class="alert alert-error">
-          <strong>Error:</strong>
-          Exception raised by {{ app.plugin.name }}: "{{ app.error }}"
-        </div>
-      {% endif %}
+    <form id="sodar-ajax-search" name="sodar-ajax-search" action={% url 'projectroles:ajax_search' %} style="display: none;">
+      <textarea name="terms">{% render_json search_terms %}</textarea>
```
mikkonie: There is also the built-in Django filter `json_script`, maybe that would work here?

https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#json-script

projectroles/plugins.py
```diff
@@ -939,37 +938,60 @@ def __init__(self, url: str, name: str, blank: bool = False):
...
+    """
+    Represents a cell in the table of search results returned by app plugins.
+    """
+
+    value: str = None
+    value_url: str = None
+    value_url_class: str = None
+    icons: list[dict[str, str]] = None
+    snippets: list[str] = None
+    cell_class: str = None
```
mikkonie: Specifying the class for each row cell is redundant, except for cases where we want to append classes to a specific row or cell.

I'd like to have this defined in the corresponding column object (see my other comment regarding column definitions) and then used for all the row cells in that column.

We should leave this in as an optional arg for the `PluginSearchResultCell` class. If a value is present, the JQuery should append given classes to the classes we get from the column object. Also, we can rename this arg into simply `class`.

projectroles/plugins.py
```diff
@@ -939,37 +938,60 @@ def __init__(self, url: str, name: str, blank: bool = False):
...
+class PluginSearchResultCell:
+    """
+    Represents a cell in the table of search results returned by app plugins.
+    """
+
+    value: str = None
+    value_url: str = None
+    value_url_class: str = None
+    icons: list[dict[str, str]] = None
+    snippets: list[str] = None
```
mikkonie: `snippets` seems to be a very specific use cases for specific apps. It creates a bit of overhead to manage all of that in the JQuery and only works for a very specific type of thing we want to do.

While I'm generally against providing pre-created HTML from the server to the client, maybe in this case it would actually be pertient? After all, we are dealing with very specific cases with a small amount of data here.

How about having an optional arg `value_html`, which overrides `value` and `value_url` if filled? That would then be displayed with `safe` to render whatever HTML is in there.

Or, alternatively, have a boolean `value_html` field in the column object, which lets the JQuery know we should handle all `value` fields in this colum as HTML and ignore `value_url`?

Also, maybe this could also be done for `icons`? Does that warrant a specific mechanism?

We can get together to discuss an optimal solution for this.

projectroles/plugins.py
```diff
@@ -939,37 +938,60 @@ def __init__(self, url: str, name: str, blank: bool = False):
         self.blank = blank
 
 
+@dataclass
+class PluginSearchResultCell:
+    """
+    Represents a cell in the table of search results returned by app plugins.
+    """
+
+    value: str = None
```
mikkonie, line 953: These are hinted as `str` but default to `None`, I guess they should be `Optional[str]`?

projectroles/plugins.py
```diff
@@ -939,37 +938,60 @@ def __init__(self, url: str, name: str, blank: bool = False):
...
+    #: Category of the result set, used in rendering (string)
     category = None
     #: Title to be displayed for this set of search results in the UI (string)
     title = None
     #: List of one or more search type keywords for these results
     search_types = []
-    #: List or QuerySet of result objects
-    items = []
+    #: List of column titles
+    field_titles = []
```
mikkonie: Instead of a list of strings, I would prefer to have a list of `PluginSearchResultColumn` objects, which would define the following fields:

- `title`: The title to be displayed (str)
- `class`: Classes to be added to the cells for that column. (str)
- `highlight`: Whether to apply highlighting for this column (bool)

projectroles/plugins.py
```diff
@@ -939,37 +938,60 @@ def __init__(self, url: str, name: str, blank: bool = False):
...
+    #: List of field indices where search terms should be highlighted (0-based)
+    highlight_fields = []
```
mikkonie: This could instead be included in the column definitions.

projectroles/static/projectroles/js/search.js
```diff
@@ -1,67 +1,273 @@
...
+            }).text(field.value)
+          )
+      } else if (field.value !== null) {
+        td.attr('class', field.cell_class)
+          .append(field.value)
+      } else {
+        td.attr('class', 'text-muted')
+          .append('N/A')
+      }
+      if (field.icons !== null) {
```
mikkonie: The icons don't seem to be currently appearing for any search results even if set.

Image

projectroles/views_ajax.py
```diff
@@ -490,6 +500,209 @@ def post(self, request, *args, **kwargs):
...
+                            icons=icons,
+                            cell_class='sodar-overflow-container',
+                        ),
+                        PluginSearchResultCell(
+                            value=project.description,
+                            cell_class='sodar-overflow-container',
+                        ),
+                    ]
+                )
+            res = PluginSearchResult(
```
mikkonie: Project results seem to be rendered a bit weirdly at the moment. The first 5 results appear as a single column. Oddly enough, changing the pagination size or navigating to the next page makes the preceeding results appear as intended?

Image

mikkonie (changes requested):
Looking good! I have some ideas to improve the search data model, hopefully some answers to questions as well as a few possible bugs I noted.

We can get together later this week to discuss ideas and solutions if needed.

As for the two questions you had in the PR notes, I'll think about them a bit further and get back to you.

projectroles/views.py
fmarotta: My bad, fixed.

projectroles/static/projectroles/js/search.js
fmarotta: Indeed, done.

projectroles/static/projectroles/js/search.js
fmarotta: I see, in the samplesheet app. It's now supported, as an additional field in the `PluginSearchResult` object.

projectroles/views_ajax.py
fmarotta: :+1:

timeline/templatetags/timeline_tags.py
fmarotta: Yes, I introduced it as a helper for the search results, also because I didn't want to remove existing tags unnecessarily. However, since the functionality is very similar to the existing `get_app_icon_html()` (it mostly just wraps the icon inside a `<span>` badge), and since `get_app_icon_html()` is only used in search results and `_list_item.html`, we could maybe also consider replacing `get_app_icon_html()` with `get_app_badge()`, which is more general. A side effect would be that the link applies to the text, not just to the icon, which could make it more consistent with the other badges, but let me know what do you think.

projectroles/plugins.py
fmarotta: :+1:

projectroles/templates/projectroles/search_results.html
fmarotta: There is, but it is not completely equivalent since it wraps the JSON inside its own `<script>` tag. I think a `render_json` tag is more straightforward, but I'm open to using `json_script` if you'd rather avoid introducing a new ad-hoc tag.

projectroles/plugins.py
fmarotta: Done as discussed IRL. However, we can't just rename it to `class` because Python doesn't allow using a reserved word as a field name. Now we have `column_class` and `cell_class`, but we could call it something else, if you prefer.

projectroles/plugins.py
fmarotta: Done (see also the previous comment).

projectroles/plugins.py
fmarotta: Done as discussed IRL.

projectroles/plugins.py
fmarotta: Done as part of the previous point.

projectroles/static/projectroles/js/search.js
fmarotta: Should be fixed now. For filesfolders, the icons are shown only for public link or flags, unlike in the `_list_item.html` template where there is also an icon for the type (file, folder, or link).

projectroles/views_ajax.py
fmarotta: Interesting, it didn't look like this on my end. Is it fixed now?

fmarotta:
Thanks for the review, I have addressed the comments and modified the implementation as discussed IRL. As regards the CSS, I've added an explicit field to the plugin objects with the path to the stylesheet, as discussed. I have also made two changes which could deserve some extra discussion (I have edited the PR summary, but I'll copy the changes here for convenience):

> Changes to template tags:
> * Replace `timeline.get_app_icon_html()` with `timeline.get_app_badge()` which is more general and can be used both in search results and in `_list_item.html`. As a side effect, the link applies to the text, not just the icon, which should make this badge more consistent with the others.
> 
> Changes to CSS classes:
> * `.sodar-overflow-container` should now be applied to the parent `<TD>` tag instead of the inner `<DIV>`. The CSS and JS work under this assumption now. The reason is that the timeline plugin search table has a dynamic cell class which changes the background color based on the timeline event status; this should be applied to the `<TD>`. It seemed more economical to assign the dynamic class to the whole `<TD>`, but let me know if you have better ideas.

projectroles/templates/projectroles/search_results.html
mikkonie: Since the current implementation requires us to wrap this in a form and a textarea object anyway, the Django best practice of using `json_script` and accessing data via `.textContent` would IMHO be preferrable.

projectroles/plugins.py
mikkonie: Oh, good point! I've been working with Javascript for too long. The naming is fine then :)

projectroles/static/projectroles/js/search.js
mikkonie: Oh yes, sorry about that. I thought I had those set up in my example data but apparently didn't. My bad.

projectroles/views_ajax.py
mikkonie: It seems to be fixed. Not sure why that was happening, I did a force reload and it didn't help, so it shouldn't be e.g. a cached CSS issue either..

timeline/templatetags/timeline_tags.py
mikkonie: Yes, that seems like the logical way to go. Anything that reduces repetition and isn't overly complicated is fine with me :)

filesfolders/plugins.py
```diff
@@ -84,8 +90,8 @@ class ProjectAppPlugin(ProjectAppPluginPoint):
     #: List of search object types for the app
     search_types = ['file', 'folder', 'link']
 
-    #: Search results template
-    search_template = 'filesfolders/_search_results.html'
+    #: Search results CSS
+    search_css = 'filesfolders/css/search.css'
```
mikkonie, line 94: Make sure to document this in `dev_project_app.rst` under "ProjectAppPlugin".

Also, references to `search_template` can be removed :)

filesfolders/plugins.py
```diff
@@ -195,17 +200,124 @@ def search(
...
-                for x in items
-                if user.has_perm('filesfolders.view_data', x.project)
-            ]
+        else:
+            items = []
+        rows = []
+        for item in items:
+            if not user.has_perm('filesfolders.view_data', item.project):
+                continue
+            if get_class(item) == 'HyperLink':
```
mikkonie: Since this is not template code, we can simply access the class name with `item.__class__.__name__`. Also, it could be called just once and stored into a local variable for brevity, instead of being called for the same item for every if/elif.

filesfolders/plugins.py
```diff
@@ -195,17 +200,124 @@ def search(
...
+                PluginSearchResultColumn(
+                    title='Name',
+                    column_class='sodar-overflow-container',
+                    highlight=True,
+                    value_html=True,
+                ),
+                PluginSearchResultColumn(
+                    title='Type',
+                    column_class='text-nowrap',
+                    highlight=False,
```
mikkonie: Since the defaults for both `highlight` and `value_html` is `False`, we can omit the kwargs unless we specifically want to set them `True`. That will keep the code more compact and cleaner.

filesfolders/plugins.py
```diff
@@ -195,17 +200,124 @@ def search(
...
         ret = PluginSearchResult(
             category='all',
             title='Files, Folders and Links',
+            table_id='sodar-ff-search-table',
             search_types=['file', 'folder', 'link'],
-            items=items,
+            columns=[
+                PluginSearchResultColumn(
+                    title='Name',
+                    column_class='sodar-overflow-container',
```
mikkonie: As discussed before, I'd prefer to have an optional `overflow` boolean argument for the column. This way we can control how overflow works even if we change the internal implementation in SODAR Core.

If the `overflow` arg is set to `True`, the related JQuery should apply the `sodar-overflow-container` class to cells in this column.

Standard Bootstrap classes such as `text-nowrap` and `text-right` are OK to add as it's currently done, as they are not internal SODAR Core stuff.

projectroles/static/projectroles/css/projectroles.css
```diff
@@ -308,7 +308,7 @@ table.sodar-card-table-bordered tbody tr:last-child td {
 }
 
 /* Container for overflowing table data */
-.sodar-overflow-container {
+.sodar-overflow-container div {
```
mikkonie: I'm not sure this change is a good idea, as we may not have a nested `<div>` element inside every instance of an element with the `.sodar-overflow-container` class. In fact, there are multiple cases in the SODAR Server repo where this is not the case. This seems to break existing implementations.

filesfolders/plugins.py
```diff
@@ -195,17 +200,124 @@ def search(
...
+                get_class(item) == 'File'
+                and item.public_url
+                and allow_public_links
+                and can_share_link
+            ):
+                share_url = reverse(
+                    'filesfolders:file_public_link',
+                    kwargs={'file': item.sodar_uuid},
+                )
+                name_value += f' <a href="{share_url}" title="Public link"><i class="iconify" data-icon="mdi:link-variant"></i></a>'
```
mikkonie: As a general note, make sure line lengths don't exceed 80 characters. These can be cleaned up once we're otherwise done with the PR of course. I guess ruff doesn't automatically wrap these? (Black had similar issues with string literals IIRC)

projectroles/static/projectroles/js/projectroles.js
```diff
@@ -131,7 +131,7 @@ $(document).ready(function () {
 
 
 function modifyCellOverflow() {
-  $('.sodar-overflow-container').each(function () {
+  $('.sodar-overflow-container div').each(function () {
```
mikkonie: Same comment as I had earlier regarding `.sodar-overflow-container`, this will break existing functionality. If there are some differing implementations here, we'll have to figure out how to support both. Maybe there needs to be a separate overflow mechanism speficifally for search results.

projectroles/templates/projectroles/_remote_project_list_item.html
```diff
@@ -4,8 +4,8 @@
 {% get_remote_project_obj site project as remote_project %}
 
 <tr>
-  <td class="align-middle">
-    <div class="sodar-overflow-container">
+  <td class="align-middle sodar-overflow-container">
+    <div>
```
mikkonie: As written in other comments, we shouldn't change `sodar-overflow-container` behaviour as that will break things in SODAR Core using projects.

projectroles/plugins.py
```diff
@@ -939,46 +938,105 @@ def __init__(self, url: str, name: str, blank: bool = False):
         self.blank = blank
 
 
+@dataclass
+class PluginSearchResultCell:
+    """
+    Represents a cell in the table of search results returned by app plugins.
+    """
+
+    #: XXX
```
mikkonie: These placeholders probably need to be replaced :)

projectroles/plugins.py
```diff
@@ -939,46 +938,105 @@ def __init__(self, url: str, name: str, blank: bool = False):
...
+    """
+
+    #: The title to be displayed (str)
+    title: str
+    #: Classes to be added to the cells for that column (str)
+    column_class: Optional[str] = None
+    #: Whether to apply highlighting for this column (bool)
+    highlight: bool = False
+    #: Interpret the value field of PluginSearchResultCell as HTML (bool)
+    value_html: bool = False
```
mikkonie, line 977: One thing I forgot to think about in my original spec is that we could also allow controlling whether a specific column can be ordered or filtered in DataTables.

I added that as a separate issue: https://github.com/bihealth/sodar-core/issues/1927

That can either be done together with this PR, or separately once we are done with this one. It's your call.

projectroles/plugins.py
```diff
@@ -939,46 +938,105 @@ def __init__(self, url: str, name: str, blank: bool = False):
...
     #: List of one or more search type keywords for these results
     search_types = []
-    #: List or QuerySet of result objects
-    items = []
+    #: List of columns to be shown in the results table
+    columns = []
+    #: List of lists of result objects
+    rows = []
+    #: Limit for the number of items returned (if <0, no limit is applied)
+    result_limit = -1
```
mikkonie: Could this also be `0`? That would be consistent with other parts of implementation where we use `0` for "no limit". I don't think there is a use case for limiting results specifically to 0, right? :)

projectroles/plugins.py
```diff
@@ -939,46 +938,105 @@ def __init__(self, url: str, name: str, blank: bool = False):
...
 
     def __init__(
         self,
         category: str,
         title: str,
-        search_types: Optional[list[str]],
-        items: Union[list, QuerySet],
+        table_id: str,
```
mikkonie: I'm thinking, should we make the `table_id` optional? It's only used for CSS styling, right?

We could have projectroles construct the table IDs automatically from the plugin name and category name, but allow overriding with a specific `table_id` if necessary.

We can discuss this in person.

projectroles/templates/projectroles/search_results.html
```diff
@@ -96,56 +47,24 @@ <h2 class="sodar-pr-content-title pt-auto">Search Results</h2>
...
+      <textarea name="keywords">{% render_json search_keywords %}</textarea>
+    </form>
+    {% for app in search_apps %}
+      <div class="sodar-ajax-search-results" data-app-name="{{ app.name }}" data-app-title="{{ app.title }}" data-app-icon="{{ app.icon }}">
+        <i class="iconify spin text-secondary"
```
mikkonie: As a small cosmetic thing, I'd like to adjust the spinner as follows:

- Only have one spinner, which is placed below search results until all of our Ajax requests have returned with a response
- Center the spinner in the viewport (should be doable by wrapping it in a div with the `w-100` Bootstrap class)

projectroles/static/projectroles/js/search.js
```diff
@@ -1,67 +1,270 @@
...
+  const tbody = $('<tbody>')
+  for (let row of result.rows) {
+    const tr = $('<tr>')
+    for (let fieldIdx in row) {
+      const column = result.columns[fieldIdx]
+      const cell = row[fieldIdx]
+      const cellDiv = $('<div>')
+      if (column.value_html == true) {
+        cellDiv.append(cell.value)
+      } else if (cell.value_url !== null) {
```
mikkonie: The `!== null` comparisons can probably be simplified into e.g. `if (cell.value_url)`, unless we specifically need to compare between empty string and null.

timeline/plugins.py
```diff
@@ -132,11 +148,62 @@ def search(
                 TimelineEvent.objects.find(search_terms, projects, kwargs)
             )
             items = [e for e in events if self._check_permission(user, e)]
+        plugin_lookup = get_plugin_lookup()
+        rows = []
+        for item in items:
+            app_badge = get_app_badge(
+                item, plugin_lookup, extra_class='badge-secondary mr-1'
```
mikkonie: For some reason, the margins between badges and the description look different between the search results and other timeline views. Maybe there is some other CSS than `mr-1` affecting this?

projectroles/templates/projectroles/search_results.html
```diff
@@ -20,59 +20,10 @@
   <!-- DataTables -->
   {% include 'projectroles/_datatables_css.html' %}
 
-  <style type="text/css">
```
mikkonie, line 23: One interesting bug I discovered: When using the `type` keyword corresponding to a plugin not in the projectroles app, the results CSS breaks. Maybe some CSS include is accidentally left out if projectroles results are not returned?

Example:

Image

mikkonie (changes requested):
Solid work, again. I was very nitpicky in my review so we'll get everything right, but all in all, this is working very nicely!

I identified one bug and had one further feature request. The latter can be put in this PR or later in a separate one, you can make the call there. The rest of my comments is mostly cleanup stuff.

Note that I mostly skipped reviewing tests, as those are still under work anyway.

filesfolders/plugins.py
fmarotta: :+1:

filesfolders/plugins.py
fmarotta: Done.

filesfolders/plugins.py
fmarotta: Done.

filesfolders/plugins.py
fmarotta: Good idea, I had forgotten about that! This should also solve some of the issues below.

projectroles/static/projectroles/css/projectroles.css
fmarotta: I agree, it's better to have the `overflow` attribute in `PluginSearchResultColumn` as you suggested.

filesfolders/plugins.py
fmarotta: Right, I over-relied on the automatic formatting and forgot to check manually.

projectroles/static/projectroles/js/projectroles.js
fmarotta: I have reverted this change, as the problem should be fixed by adding the `overflow` column attribute.

projectroles/templates/projectroles/_remote_project_list_item.html
fmarotta: :+1:

projectroles/plugins.py
fmarotta: Good catch.

projectroles/plugins.py
fmarotta: Nice idea. I'd prefer to do it separately.

projectroles/plugins.py
fmarotta: Sure! :)

projectroles/templates/projectroles/search_results.html
fmarotta: Done.

projectroles/static/projectroles/js/search.js
fmarotta: Done. In the case of `cell.value`, I think we do want to distinguish between empty string and null, so I left it as it was, but let me know if you think otherwise.

projectroles/templates/projectroles/search_results.html
fmarotta: Good catch, it was my bad.

timeline/plugins.py
fmarotta: Another good catch. Should be fixed now.

projectroles/plugins.py
fmarotta: That could work, but I need to wrap my head around the implementation.. Maybe we should discuss in person.

projectroles/static/projectroles/js/search.js
mikkonie: Fair enough, in that case it's of course valid to have that comparison :)

docs/source/dev_project_app.rst
```diff
@@ -243,8 +243,9 @@ Implementing the following is **optional**:
     definition for an example.
 ``search_types``
     Implement if searching the data of the app is enabled.
-``search_template``
-    Implement if searching the data of the app is enabled.
+``search_css``
+    Path to CSS file styling the search results for this plugin. Implement if
+    searching the data of the app is enabled.
```
mikkonie: We could clarify that this should be set if CSS needs to be customized in the plugin search results. This should be optional, after all.

projectroles/templates/projectroles/search_results.html
```diff
@@ -96,56 +52,24 @@ <h2 class="sodar-pr-content-title pt-auto">Search Results</h2>
...
-      {% elif app.error %}
-        <div class="alert alert-error">
-          <strong>Error:</strong>
-          Exception raised by {{ app.plugin.name }}: "{{ app.error }}"
-        </div>
-      {% endif %}
+    {{ search_terms|json_script:"search-terms" }}
+    {{ search_keywords|json_script:"search-keywords" }}
+    {% for app in search_apps %}
+      <div class="sodar-ajax-search-results" data-url="{% url 'projectroles:ajax_search' %}" data-app-name="{{ app.name }}" data-app-title="{{ app.title }}" data-app-icon="{{ app.icon }}">
```
mikkonie: This long line could be broken down for readability into e.g.:
```
<div class="sodar-ajax-search-results"
     data-url="{% url 'projectroles:ajax_search' %}"
     data-app-name="{{ app.name }}"
     data-app-title="{{ app.title }}"
     data-app-icon="{{ app.icon }}">
```

projectroles/templates/projectroles/search_results.html
```diff
@@ -96,56 +52,24 @@ <h2 class="sodar-pr-content-title pt-auto">Search Results</h2>
...
-          <li>{{ n }}</li>
-        {% endfor %}
-      </ul>
+    <div id="sodar-search-results-spinner" class="w-100 text-center">
+      <i class="iconify spin text-secondary"
+         data-icon="mdi:loading" data-height="42">
+      </i>
     </div>
   {% endif %}
+
```
mikkonie: This blank line we can probably omit :)

projectroles/plugins.py
```diff
@@ -939,46 +938,114 @@ def __init__(self, url: str, name: str, blank: bool = False):
         self.blank = blank
 
 
+@dataclass
```
mikkonie, line 942: No change to request here, just wanted to say that it's a good call to use the `dataclasses` module here. I guess we should also do that for other existing data-specific classes where applicable. Anyway, that's for another PR :)

timeline/plugins.py
```diff
@@ -132,11 +148,56 @@ def search(
...
+            rows.append(
+                [
+                    # Timestamp
+                    PluginSearchResultCell(
+                        value=get_timestamp(item),
+                    ),
+                    # Description
+                    PluginSearchResultCell(
+                        value=(
+                            app_badge
```
mikkonie: Maybe we should construct the value in a variable before `rows.append()` so the formatting will look a bit nicer.

timeline/plugins.py
```diff
@@ -132,11 +148,56 @@ def search(
                 TimelineEvent.objects.find(search_terms, projects, kwargs)
             )
             items = [e for e in events if self._check_permission(user, e)]
+        plugin_lookup = get_plugin_lookup()
+        rows = []
+        for item in items:
+            app_badge = get_app_badge(item, plugin_lookup, extra_class='mr-1')
+            user_badge = get_user_badge(item.user, extra_class='mr-1')
+            project_badge = get_project_badge(item.project, extra_class='mr-1')
+            rows.append(
```
mikkonie, line 173: The margins between badges and description test still appear inconsistent between search results and timeline app views. It would be nice if these would be identical.

Search results:

Image

Timeline app view (expected layout):

Image

mikkonie (changes requested):
Good work once again. Everything seems to work nicely, I only had a couple of minor cleanup/docs/cosmetic notes.

As stated in the previous review, I'd still like the `table_id` for search results to be optional. Hit me up and we can discuss this.

mikkonie:
I thought about it and the `table_id` feature should work as follows:

- Replace `table_id` with `table_class`, which is `Optional[str]`
- If the table class string is set, append it to the related results table class in JQuery
- Refer to table class instead of ID in custom search CSS

This is easy to implement and flexible: it allows specifying the same CSS for multiple results tables from the same plugin if needed, as ID elements are supposed to be unique.

If there are any questions on this, let's talk.

projectroles/templates/projectroles/search_results.html
fmarotta: Done

projectroles/templates/projectroles/search_results.html
fmarotta: Fixed

projectroles/plugins.py
fmarotta: Hehe, feel free to assign that to me :)

timeline/plugins.py
fmarotta: :+1:

timeline/plugins.py
fmarotta: Good catch. The `timeline/_list_item.html` template sneakily introduces spaces between the badges. I added them to the search results as well.

docs/source/dev_project_app.rst
fmarotta: Updated

fmarotta (commented):
I have addressed the comments and now the PR is ready for another review, sorry for taking a long time. Here is a summary of the main changes since last time:
- Replace `table_id` with `table_class` in `PluginSearchResult` and update CSS
- Update display of timeline search results to make it consistent with list view
- Add `get_event_full_description()` function in `timeline/plugin.py`
- Fix search results for PROJECT_FINDER role (I found a bug)
- Added tests
- Fixed shepherdjs tour
- Updated docs

fmarotta:
I just realized that I introduced the fix for #1926 here as well.

mikkonie (approved):
Good work! I can't find any issues worth another round of revisions at this point. The tests look thorough and I did my best to break things in dev without success :)

Some notes I had which we might consider in future PRs, but IMHO not worth tackling right now:

- Some `search()` functions are quite lengthy due to unavoidable complexity needed for providing result metadata to the client. Maybe we might want to look into splitting these into internal helpers at some point.
- I noticed some RST syntax in class docstrings for the data classes. Generally, I want to avoid this except for REST API views where we really need the extra formatting. However, this might be a me thing and I need to think about it more. Leaving them in for now.

Thank you for the hard work on this issue, which turned out to be way more involved than I initially envisioned :)

Issue #1923, opened 2026-04-13
Add pseudonymization option in removeroles command


mikkonie:
While discussing GDPR issues for public-facing SODAR instances, there was talk about user account removal and user data retaining.

Currently, if we decativate a user via `removeroles` (or the Django admin), their information is no longer visible to end users. Alas, we still do store their real name and email address on the server.

To support removing personal data, we should add an option to `removeroles` which clears out all personal user data from their `User` object. To offer maximum pseudonymity, we can also replace the username with a random string.

## Spec

- Add argument to `removeroles`
   * Args = `-p` or `--pseudonymize`
       + (Or something better if you have ideas)
   * Use `action='store_true'` so no value needed
   * Mention in help that this is only applied if `--deactivate` is used
- Add pseudonymization logic to `handle()`
   * Only if `deactivate` is set
   * Set `full_name`, `first_name` and `last_name` to empty strings
   * Set `email` to an empty string
   * Replace user name with `deactivated` + random string of lowercase letters and numbers
       + Make sure generated user name doesn't exist already
- Add/update tests

## Tasks

- [ ] Add argument
- [ ] Add logic in command
- [ ] Update tests

mikkonie:
Done by @fmarotta, merged.

PR #1929, opened 2026-05-13
Add pseudonymization option to removeroles command

fmarotta:
This PR addresses #1923. There is one departure from the original spec: I used `name` instead of `full_name` because the latter doesn't exist (which is confusing, because there is a `get_full_name()`).

While working on it, I noticed that users that have no roles are not deactivated (due to an early return statement). Thinking that we may want to pseudonymize them even if they have no roles, I went ahead and removed the return statement, but please let me know if we should update the logic.

### DIFF
--- a/projectroles/management/commands/removeroles.py
+++ b/projectroles/management/commands/removeroles.py
@@ -145,7 +145,7 @@ class Command(
                 user.last_name = ''
                 user.email = ''
                 while True:
-                    random_username = 'deactivated_' + ''.join(
+                    random_username = 'deactivated' + ''.join(
                         random.choices(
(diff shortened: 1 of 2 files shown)


mikkonie:
> This PR addresses https://github.com/bihealth/sodar-core/issues/1923. There is one departure from the original spec: I used name instead of full_name because the latter doesn't exist (which is confusing, because there is a get_full_name()).

Oh yes, good call. This should be `name` of course, that was my bad.

The `get_full_name()` is logical, because it either returns `name` or builds a full name from `first_name` and `last_name`, as these fields may be differently filled depending on app, site or third-party backend.

mikkonie (approved):
No notes, looking good!

Issue #1927, opened 2026-05-12
Allow disabling  search result column ordering and filtering


mikkonie:
Something I thought of while reviewing #22.

DataTables allows us to disable ordering and/or searching (meaning the filtering of table content) for specific columns in a table. It would be nice if we can do this for the async search results.

For example: it would be a good idea to disable searching for the item type column in filesfolders results, as we can already limit the search items with a keyword. Similarly, we might have some results where ordering contant by a specific column does not make sense, if it e.g. contains just some icons or similar not-very-orderable content.

Hence I would propose the following..

## Spec

- Add `orderable` and `searchable` members in `PluginSearchResultColumn`
    * Default to `True`
- When building a DataTable for search results in JQuery, add `columnDefs` according to values in the columns
    * For an example, [see role_list.js](https://github.com/bihealth/sodar-core/blob/88e1c9bc91dbe8c2f7771a8ef9f4b933d3c66287/projectroles/static/projectroles/js/role_list.js#L16)
- Add UI tests to verify this
- Update docs regarding these optional args

This can either be implemented with the PR for #22, or separately after it. This should not break anything in the new search implementation, just add an optional feature. However, if you'd just like to get it done with in the same PR, I'm OK with that too.

mikkonie:
Done by @fmarotta, merged.

PR #1944, opened 2026-06-11
Disable search result column ordering and filtering

fmarotta:
This PR addresses #1927 and one of the notes at the end of the review of #1929. Among other things, I've added a utility function to simplify the filesfolder search results, split the `get_search_results()` function in two parts, and refactored the search.js code into smaller functions.


filesfolders/tests/test_ui.py
```diff
@@ -475,6 +476,49 @@ def _assert_search_results_count(self, expected, url):
             )
             self.assertEqual(len(elements), expected_count)
 
+    def _assert_search_results_filtering_count(
```
mikkonie: We define a rather lengthy internal helper function three times, the only difference appearing to be the table class and app name.

How about adding this as a single generic function under a mixin in `projectroles/tests/base.py` and import that to tests when necessary? We could call that mixin, say, `SearchUITestMixin`.

Also a minor note on styling: let's try to keep function names as concise as possible. Generally, conjugating words with `*ing` is a bit redundant. Also some bits we already know from context, so e.g. this function could be named `assert_search_filter_count()`.

filesfolders/utils.py
```diff
@@ -20,3 +28,73 @@ def build_public_url(file: File, request: HttpRequest) -> str:
             kwargs={'secret': file.secret, 'file_name': file.name},
         )
     )
+
+
+def extract_properties(
```
mikkonie: If we only ever call this from plugins, we could make this an internal helper function within `Filesfolders.plugins.ProjectAppPlugin`. That way we reduce the complexity of `search()` while still maintaining functions in a logical place.

Furthermore, the naming of this function seems somewhat vague. "Extract properties" does not really tell us much about what we're doing. Maybe something like `_get_search_properties()`?

projectroles/plugins.py
```diff
@@ -971,6 +977,10 @@ class PluginSearchResultColumn:
     value_html: bool = False
     #: Whether the column content is allowed to overflow (bool)
     overflow: bool = False
+    #: Whether users can order the table by this column's values (bool)
+    orderable: bool = True
```
mikkonie, line 975: Make sure these are documented in `dev_project_app.rst` where discussing `PluginSearchResultColumn`.

timeline/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    get_project_badge,
+)
+
+from timeline.templatetags.timeline_tags import (
+    get_app_badge,
+    get_event_description,
+)
+
+
+def get_event_full_description(event, plugin_lookup, extra_class='mr-1'):
```
mikkonie: Here we have the same issue as in filesfolders: We define an utility function which is only called from the plugin, so it could be done as an internal helper in the plugin class.

However, I see you're also calling this directly in tests. Maybe that makes it make sense to have it as a separate utility method. I'd say you can make the call on this one.

mikkonie (changes requested):
Good work, seems to be working without issues. I had some suggestions for refactoring and documentation.

filesfolders/tests/test_ui.py
fmarotta: Good ideas, done.

filesfolders/utils.py
fmarotta: Yes, so far it's only used for the search table, so I agree, it makes sense to call it `_get_search_properties()`.

projectroles/plugins.py
fmarotta: My bad, fixed.

timeline/utils.py
fmarotta: I moved it to the plugin class, it seems more consistent with what we do in filesfolders.

mikkonie (approved):
Looks great, no notes!

PR #1956, opened 2026-07-17
Correctly set old_role in modify_assignment()

### LINKED ISSUE #1955: RoleAssignmentModifyMixin.modify_assignment() sets old_role to a wrong value
During the SODAR migration to SODAR Core v1.4, some tests were failing because of this. PR #1875 introduced a [bug](https://github.com/bihealth/sodar-core/pull/1875/changes#diff-788ba64cb0c4a9cac4e25001258d75ffe41fcbf5c65b01242fc4b9d28ba7d33aL2160-R2163) in `RoleAssignmentModifyMixin.modify_assignment()`: I misinterpreted the `instance` parameter to be the "old" instance to be updated, but actually it can be the newly instantiated object if it comes from from the form in `RoleAssignmentUpdateView`. This means that, in that case, the `old_role` variable gets the wrong value.

fmarotta:
This PR attempts to fix #1955. Instead of wrongly assuming that `instance` is the old role assignment, we try to get the actual `RoleAssignment` object, falling back to `project.get_role(user)` if the `get()` query fails (which can happen if the role is inherited).

As a regression test, I implemented the `perform_role_modify()` method in the `example_project_app` plugin. It sets a message so that we can check if `old_role` is assigned correctly.

### DIFF
--- a/projectroles/tests/test_views.py
+++ b/projectroles/tests/test_views.py
@@ -90,6 +90,7 @@ from projectroles.views import (
     PROJECT_DELETE_CAT_ERR_MSG,
     PROJECT_DELETE_SOURCE_ERR_MSG,
     PROJECT_DELETE_TARGET_ERR_MSG,
+    PROJECT_ACTION_UPDATE,
 )
 from projectroles.context_processors import (
     SIDEBAR_ICON_MIN_SIZE,
@@ -4475,7 +4476,7 @@ class TestRoleAssignmentUpdateView(
         self.assertEqual(
             list(get_messages(response.wsgi_request))[0].message,
             EXAMPLE_ROLE_MODIFY_API_MSG.format(
-                action='update',
+                action=PROJECT_ACTION_UPDATE,
                 old_role=old_role,
                 new_role=self.role_contributor,
                 user=self.role_as.user,
--- a/projectroles/tests/test_views.py
+++ b/projectroles/tests/test_views.py
@@ -22,6 +22,8 @@ from test_plus.test import TestCase
 # Timeline dependency
 from timeline.models import TimelineEvent
 
+from example_project_app.plugins import EXAMPLE_ROLE_MODIFY_API_MSG
+
 from projectroles.app_settings import AppSettingAPI
 from projectroles.email import (
     get_email_user,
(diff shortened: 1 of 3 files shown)


projectroles/tests/test_views.py
```diff
@@ -4453,6 +4455,34 @@ def test_post(self):
...
+                response,
+                reverse(
+                    'projectroles:roles',
+                    kwargs={'project': self.project.sodar_uuid},
+                ),
+            )
+        self.assertEqual(
+            list(get_messages(response.wsgi_request))[0].message,
+            EXAMPLE_ROLE_MODIFY_API_MSG.format(
+                action='update',
```
mikkonie: I recommend using the `PROJECT_ACTION_UPDATE` constant here just to make things consistent.

projectroles/views.py
```diff
@@ -2096,7 +2096,10 @@ def modify_assignment(
             role_as = RoleAssignment(project=project, user=user, role=role)
             old_role = None
         else:
-            role_as = instance
+            try:
+                role_as = RoleAssignment.objects.get(project=project, user=user)
+            except RoleAssignment.DoesNotExist:
+                role_as = project.get_role(user)
```
mikkonie: The original line of `role_as = instance` is correct. If we call this function with the update action, the `RoleAssignment` object to be updated is that specific one. If we're calling update with a different object, something is wrong with the caller.

A problem with this fix is that when retrieving roles with `Project.get_role()`, we may end up with an inherited `RoleAssigmnent` object and we then proceed to modify it here. We don't want to do that.

However, it is true that `old_role` for calling the project modify API can not directly be derived from the `instance` object in case of inheritance. Good catch.

I believe this should work something like this:

```
else:
    role_as = instance
    old_as = project.get_role(user)
    if old_as and old_as != role_as:
        old_role = old_as.role
    else:
        old_role = role_as.role
    role_as.role = role
role_as.save()

# Then continue as before..
```

It's possible I missed something though, or maybe there's an even better way to do this. We can discuss this once back at the office.

mikkonie (changes requested):
Thanks for the fix. I have a suggestion for a slightly different approach, we can discuss this IRL.

projectroles/tests/test_views.py
fmarotta: Done

projectroles/views.py
fmarotta: Good point, I fixed the fix.

mikkonie (approved):
Looks good, merging :)

PR #1960, opened 2026-07-20
do not show "undefined" if info_link is missing (#1959)

### LINKED ISSUE #1959: Site info labels show "undefined" if the info_link is missing
A bug which I introduced in the site info page while working on #1880, obviously the "undefined" shouldn't be there:

Image

fmarotta:
This PR attempts to fix #1959.


mikkonie (approved):
Approved, well spotted. I had totally missed this as well :)

Issue #1958, opened 2026-07-20
Add HostConfirmDeleteView optional permission check


mikkonie:
As spoke @fmarotta in https://github.com/bihealth/sodar-server/pull/2483:

> In sodar-server, the form is not shown if can_delete_sheets is false. This is not taken into account in the SODAR Core template, so I had to duplicate the entire template instead of extending it.

This is a good point and I didn't think of it while this feature was being implemented. I think the way this is done in SODAR is valid, as we want to present the user with info on why they are not allowed to do this action.

My suggestion on how to handle this in `HostConfirmDeleteView`:

- In the template, check for a specific context variable
   * We can call it e.g. `allow_delete`
   * Assumed to be boolean, can either be set or not
       + As we know, Django templates don't care if vars are unset, so we don't have to worry about that :)
- If context variable is set but not `True`, hide the host confirmation field and only display the cancel button

This should be nice and simple, but if you have a better approach in mind, I'm all ears.

This is a minor non-breaking release, so it would go nicely into v1.4.2.

mikkonie:
Done by @fmarotta, merged.

PR #1961, opened 2026-07-20
Add permission check in HostConfirmDeleteView template

fmarotta:
This PR attempts to address #1958.

### DIFF
--- a/projectroles/templates/projectroles/confirm_delete_host.html
+++ b/projectroles/templates/projectroles/confirm_delete_host.html
@@ -23,20 +23,26 @@
 
   <form method="post">
     {% csrf_token %}
-    <input class="form-control mb-3"
-           name="delete_host_confirm"
-           autocomplete="off"
-           required="true"
-           placeholder="Type the host name of this server (e.g. &quot;site.example.com&quot;) to confirm deletion." />
-    <div class="btn-group pull-right">
-      <a role="button" class="btn btn-secondary"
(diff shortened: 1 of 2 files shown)


projectroles/templates/projectroles/confirm_delete_host.html
```diff
@@ -23,20 +23,26 @@ <h3>Confirm Deletion of {{ object_display_name|title }}<h3>
...
-      <a role="button" class="btn btn-secondary"
-         href="{{ request.session.real_referer }}">
-        <i class="iconify" data-icon="mdi:arrow-left-circle"></i> Cancel
-      </a>
-      <button type="submit" class="btn btn-danger sodar-btn-submit-once"
-              id="sodar-pr-btn-confirm-delete">
-        <i class="iconify" data-icon="mdi:close-thick"></i> Delete
-      </button>
+    <div class="row">
+      {% if allow_delete is None or allow_delete is True %}
```
mikkonie: This could probably be simplified to `{% if allow_delete is not False %}`?

mikkonie (changes requested):
Very quick work! :) I have one small suggestion.

projectroles/templates/projectroles/confirm_delete_host.html
fmarotta: Good point, updated!

mikkonie (approved):
Looking good, accepted!

Issue #1971, opened 2026-08-10
Incorrect app alert badge content alignment


mikkonie:
After doing #1967 my eyes must be sensitive to this stuff.

The vertical alignment of content inside `sodar-app-alert-badge` is incorrect. See text vs the close button.

Concerns both Firefox and Blink-based browsers.

Not urgent, something for a quiet Friday afternoon.

mikkonie:
Spent some time tweaking this, gave up for now. Very hard to get it to appear right on both major browser platform without browser-specific attributes I'd prefer to avoid.

I'll return to this when I'm doing some other CSS stuff. Or in v2.0 at the latest.

PR #1974, opened 2026-08-13
Fix search results highlight for paginated tables

fmarotta:
This PR addresses #1971. After initializing the `DataTable`, the table rows in page 2 and subsequent are removed from the DOM, so the jQuery selector used to find search terms could not find those rows. The same problem affected the `modifyCellOverflow()` function. With this PR, the calls to those functions are moved before initializing the `DataTable`, when all the cells are still in the DOM.


mikkonie (approved):
Good work, thanks for the quick fix! Merging :)

